PR #217, opened 2026-03-04
fix: ensure onboarding pre-accepted for credentials.json auth

voyax:
## Summary

Follow-up to #215. Fixes the case where a user runs `claude login` first, then installs zylos — Guardian detects the login but Claude shows onboarding/login prompt in tmux and blocks.

**Root cause:** `approveApiKey()` handled both API key approval AND onboarding setup. When using credentials.json only (no .env tokens), `approveApiKey()` was never called, so `hasCompletedOnboarding` and `hasTrustDialogAccepted` were never set in `~/.claude.json`.

**Fix:** Extract `ensureOnboardingComplete()` from `approveApiKey()` and call it unconditionally in `startClaude()` for all auth methods.

## Changes

- **`ensureOnboardingComplete()`** — new function: sets `hasCompletedOnboarding` and `hasTrustDialogAccepted` in `~/.claude.json`
- **`approveApiKey()`** — now only handles API key approval (single responsibility)
- **`startClaude()`** — calls `ensureOnboardingComplete()` unconditionally before starting Claude

## Test plan

- [ ] Fresh machine: `claude login` → install zylos → `zylos status` shows ONLINE
- [ ] Fresh machine: no login → .env with API key → zylos starts correctly
- [ ] Fresh machine: no login → .env with setup token → zylos starts correctly
- [ ] Existing machine with `~/.claude.json` already set → no changes written (idempotent)

### DIFF
--- a/skills/activity-monitor/scripts/activity-monitor.js
+++ b/skills/activity-monitor/scripts/activity-monitor.js
@@ -7,6 +7,8 @@
  *   - startClaude(): skip .env OAUTH_TOKEN injection when credentials.json is available
  *   - isClaudeLoggedIn(): check credentials.json first, before falling back to .env
  *   - Prevents 401 errors caused by expired static tokens overriding auto-refreshable ones
+ *   - Extract ensureOnboardingComplete() from approveApiKey() so onboarding/trust dialogs
+ *     are pre-accepted for ALL auth methods, including credentials.json-only installs
  *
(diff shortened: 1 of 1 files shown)


voyax:
Superseded by PR #219, which includes the ensureOnboardingComplete() extraction as part of the broader auth conflict fix.

PR #222, opened 2026-03-04
fix: only use credentials.json for native auth detection

voyax:
## Summary
- Removed `claude auth status` fallback from `hasNativeAuth` detection — only `credentials.json` counts as native auth
- On macOS, `claude auth status` reports loggedIn via system Keychain, but Keychain is inaccessible from PM2-spawned tmux sessions
- Trusting that signal caused Guardian to skip `.env` token injection, leaving tmux with no auth

## Test plan
- [ ] macOS: setup token in .env + previous `claude login` (Keychain) → `zylos attach` should work (setup token injected)
- [ ] Linux: credentials.json exists + API key in .env → no auth conflict (credentials.json = native auth, .env skipped)
- [ ] Linux: no credentials.json, only API key in .env → works (API key injected)

### DIFF
--- a/skills/activity-monitor/scripts/activity-monitor.js
+++ b/skills/activity-monitor/scripts/activity-monitor.js
@@ -460,29 +460,17 @@ function approveApiKey(apiKey) {
 }
 
 function startClaude() {
-  // Detect native `claude login` auth (credentials.json on Linux, system Keychain
-  // on macOS).  When native auth is active, .env tokens MUST NOT be injected into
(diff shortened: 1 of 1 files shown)


voyax:
Closing: the premise that PM2-spawned tmux sessions cannot access macOS Keychain was incorrect. voya.luo confirmed that `claude login` (Keychain auth) works normally inside `zylos attach` on macOS. The existing auth detection logic is correct — no change needed.

PR #236, opened 2026-03-04
fix: insert component routes into primary server block

Nanyan:
## Summary

- Fix `applyCaddyRoutes()` in `cli/lib/caddy.js` to target the first (primary) server block instead of the last one
- Add `findPrimaryBlockEnd()` helper that tracks brace depth to find the correct insertion point
- Prevents component webhook routes from being inserted into user-added server blocks

## Problem

`applyCaddyRoutes()` uses `content.lastIndexOf('}')` to find where to insert component routes. This assumes the Caddyfile has only one server block. When users add additional server blocks (e.g. reverse proxies for other services), `lastIndexOf` returns the closing brace of the **last** block, causing component routes to be inserted into the wrong host context.

**Real-world impact:** After upgrading zylos-lark (v0.1.7 → v0.1.9), the lark webhook route was inserted into a secondary `http://aidep.online` block instead of the primary `:80` block. Since Lark sends webhooks to the primary domain (`zylos305.coco.site`), all webhook requests returned 404 and the bot stopped receiving messages.

## Fix

Replace `lastIndexOf('}')` with `findPrimaryBlockEnd()` which scans from the start of the file, tracking brace depth (`{` increments, `}` decrements). When depth returns to 0, we've found the end of the first server block — always the zylos-managed one.

## Test plan

- [ ] Upgrade a component with `http_routes` on a single-block Caddyfile → route inserted correctly
- [ ] Upgrade on a multi-block Caddyfile → route inserted into first block, not last
- [ ] Nested `handle` blocks inside the primary block don't cause premature match

### DIFF
--- a/cli/lib/caddy.js
+++ b/cli/lib/caddy.js
@@ -141,19 +141,38 @@ export function switchProtocol(domain, protocol) {
 /**
  * Find the closing `}` of the first (primary) server block in a Caddyfile.
  * Tracks brace depth starting from the first `{` to find its matching `}`.
+ * Skips comment lines and Caddy global options blocks (label-less `{` at top).
  * Returns the character index of the closing brace, or -1 if not found.
  */
 function findPrimaryBlockEnd(content) {
   let depth = 0;
   let inBlock = false;
+  let skippingGlobal = false;
 
   for (let i = 0; i < content.length; i++) {
+    // Skip comment lines (# to end of line)
+    if (content[i] === '#') {
+      while (i < content.length && content[i] !== '\n') i++;
+      continue;
+    }
+
     if (content[i] === '{') {
       depth++;
-      inBlock = true;
+      if (!inBlock && !skippingGlobal) {
+        // Check if this is a global options block: no site address before {
+        const lineStart = content.lastIndexOf('\n', i) + 1;
+        const before = content.slice(lineStart, i).trim();
+        if (before === '' && depth === 1) {
+          skippingGlobal = true;
+        } else {
+          inBlock = true;
+        }
+      }
     } else if (content[i] === '}') {
       depth--;
-      if (inBlock && depth === 0) {
+      if (skippingGlobal && depth === 0) {
+        skippingGlobal = false;
+      } else if (inBlock && depth === 0) {
         return i;
       }
     }


voyax (changes requested):
Good fix for the multi-block routing issue. The brace-depth approach is clean and handles the core case well. Found 2 edge cases that should be addressed:

## Edge Case 1: Braces inside comments

```
# Config { example }
http://example.com {
    respond "hello"
}
```

`findPrimaryBlockEnd` counts `{` and `}` inside comments. In this example, the comment line's `{ }` causes depth to go 0→1→0, returning the `}` position inside the comment — wrong result.

**Risk:** Low (zylos-generated comments don't contain braces), but fragile if users add comments like `# TODO: handle { special case }`.

**Suggested fix:** Skip characters on lines starting with `#`:

```javascript
for (let i = 0; i < content.length; i++) {
  // Skip comment lines
  if (content[i] === '#') {
    while (i < content.length && content[i] !== '\n') i++;
    continue;
  }
  // ... existing brace logic
}
```

## Edge Case 2: Caddy global options block

```
{
    email admin@example.com
    acme_ca https://acme.example.com
}

http://example.com {
    respond "hello"
}
```

Caddy supports a label-less `{ }` block at the top for global options (ACME config, etc.). `findPrimaryBlockEnd` treats this as the "first server block" and returns its closing brace — routes would be inserted into the global options block.

**Risk:** Medium. zylos doesn't generate a global options block by default, but it's standard Caddy usage. Users setting up HTTPS with custom ACME config would hit this.

**Suggested fix:** Before counting, check if the `{` has a site address (hostname/port) preceding it. A bare `{` at line start (after optional whitespace) is a global options block — skip it:

```javascript
function findPrimaryBlockEnd(content) {
  let depth = 0;
  let inBlock = false;
  let isGlobalBlock = false;

  for (let i = 0; i < content.length; i++) {
    // Skip comment lines
    if (content[i] === '#') {
      while (i < content.length && content[i] !== '\n') i++;
      continue;
    }

    if (content[i] === '{') {
      depth++;
      if (!inBlock) {
        // Check if this is a global options block (no site address before {)
        const lineStart = content.lastIndexOf('\n', i) + 1;
        const before = content.slice(lineStart, i).trim();
        if (before === '' && depth === 1) {
          isGlobalBlock = true;
        } else {
          inBlock = true;
        }
      }
    } else if (content[i] === '}') {
      depth--;
      if (isGlobalBlock && depth === 0) {
        isGlobalBlock = false; // Done skipping global block
      } else if (inBlock && depth === 0) {
        return i;
      }
    }
  }

  return -1;
}
```

Both fixes are straightforward and don't change the behavior for the common case. Happy to help iterate on the implementation.

PR #296, opened 2026-03-10
fix: prevent API key exposure in process command line

### LINKED ISSUE #289: API key exposed via shell command line in activity monitor
## Summary

API keys (`ANTHROPIC_API_KEY`, `CLAUDE_CODE_OAUTH_TOKEN`) are passed as arguments to `tmux new-session -e` via `execSync()` string interpolation in the activity monitor and init scripts.

## Files

- `activity-monitor.js` (line ~642-644)
- `init.js`

## Problem

```js
const apiKeyEnv = apiKeyValue ? ` -e "ANTHROPIC_API_KEY=${apiKeyValue}"` : '';
execSync(`tmux new-session -d -s "${SESSION}" ... ${apiKeyEnv} ...`);
```

### Risks

1. **Secrets visible in process listing** — any user on the host can read them via `ps aux` or `/proc/<pid>/cmdline`
2. **Shell injection** — if `.env` values contain shell metacharacters (`;`, `` ` ``, `$()`, etc.), the string interpolation in `execSync()` creates an RCE vector

## Recommendation

- Use `execFileSync` or `child_process.spawn` with argument arrays (no shell) instead of `execSync` with string concatenation
- For tmux env vars, use `tmux setenv` after session creation instead of passing via `-e` flag

## Severity

**HIGH** — requires host access to exploit, but affects all zylos instances using API key auth.

---

Originally reported in coco-xyz/coco-dashboard#529.

voyax:
## Summary

Closes #289.

- **activity-monitor.js**: Replaced `execSync` string interpolation with `execFileSync` + argument array. Secrets (API key, OAuth token) are written to a `mode 0600` temp file, sourced inside the tmux shell command, then immediately deleted — never visible in process listing.
- **init.js**: Same pattern applied for `ANTHROPIC_API_KEY` during `guideBypassAcceptance()`.

Both shell injection (via metacharacters in `.env` values) and process listing exposure (`ps aux`, `/proc/*/cmdline`) are eliminated.

## Test plan

- [ ] Docker test: new user init with API key — verify key not in `ps aux` output
- [ ] Docker test: activity monitor restart with OAuth token — verify token not in `ps aux`
- [ ] Verify Claude starts correctly and can access the API key within the tmux session
- [ ] Verify temp env file is deleted after sourcing (check `/tmp/.zylos-env-*`)
- [ ] Edge case: no API key configured (OAuth/native auth only) — verify no temp file created

### DIFF
--- a/cli/commands/init.js
+++ b/cli/commands/init.js
@@ -723,8 +723,8 @@ async function guideBypassAcceptance() {
     let tmpEnv = null;
     if (process.env.ANTHROPIC_API_KEY) {
       tmpEnv = path.join(os.tmpdir(), `.zylos-env-${process.pid}-${Date.now()}`);
-      fs.writeFileSync(tmpEnv, `ANTHROPIC_API_KEY=${process.env.ANTHROPIC_API_KEY}\n`, { mode: 0o600 });
-      shellCmd = `set -a; . ${tmpEnv}; set +a; rm -f ${tmpEnv}; cd ${ZYLOS_DIR} && claude --dangerously-skip-permissions`;
+      fs.writeFileSync(tmpEnv, `ANTHROPIC_API_KEY='${process.env.ANTHROPIC_API_KEY}'\n`, { mode: 0o600 });
+      shellCmd = `set -a; . "${tmpEnv}"; set +a; rm -f "${tmpEnv}"; cd ${ZYLOS_DIR} && claude --dangerously-skip-permissions`;
     } else {
       shellCmd = `cd ${ZYLOS_DIR} && claude --dangerously-skip-permissions`;
     }
--- a/skills/activity-monitor/scripts/activity-monitor.js
+++ b/skills/activity-monitor/scripts/activity-monitor.js
@@ -641,15 +641,15 @@ function startClaude() {
 
       // Write secrets to a temp file, source it in the shell command, then delete it
       const envParts = [];
-      if (apiKeyValue) envParts.push(`ANTHROPIC_API_KEY=${apiKeyValue}`);
-      if (oauthTokenValue) envParts.push(`CLAUDE_CODE_OAUTH_TOKEN=${oauthTokenValue}`);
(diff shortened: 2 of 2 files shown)


voyax (commented):
## Review: PR #296

Overall approach is solid — `execFileSync` with argument array eliminates shell injection, temp env file eliminates `ps aux` / `/proc/cmdline` exposure. Two issues to address:

### Issue 1: Unquoted values in temp env file (Medium)

```js
envParts.push(`ANTHROPIC_API_KEY=${apiKeyValue}`);
```

When sourced via `. file`, if the value ever contains spaces or shell metacharacters, it breaks. API keys are typically safe (alphanumeric), but defense-in-depth says quote them:

```js
envParts.push(`ANTHROPIC_API_KEY='${apiKeyValue}'`);
```

Same for `CLAUDE_CODE_OAUTH_TOKEN` and the `ANTHROPIC_API_KEY` write in `init.js`.

(If paranoid about single quotes in values: `val.replace(/'/g, "'\\''")` — but API keys won't contain them.)

### Issue 2: Unquoted `${tmpEnv}` in shell command (Low)

```js
shellCmd = `set -a; . ${tmpEnv}; set +a; rm -f ${tmpEnv}; ...`;
```

Should quote for robustness:
```js
shellCmd = `set -a; . "${tmpEnv}"; set +a; rm -f "${tmpEnv}"; ...`;
```

`/tmp` won't have spaces on Linux, but it's free to fix and prevents surprises.

### Everything else looks good

- ✅ `execFileSync` (no shell) — prevents injection
- ✅ Temp file mode `0o600` — owner-only
- ✅ `set -a; . file; set +a; rm -f file` — correct sourcing + immediate cleanup
- ✅ Catch block deletes temp file if `execFileSync` fails
- ✅ `PID + Date.now()` in filename — prevents collisions
- ✅ `sendToTmux` path (existing session restart) is unaffected — env vars persist in tmux session env from initial creation
- ✅ Native auth path (`hasNativeAuth`) correctly skips secret injection
- ✅ `init.js` only handles `ANTHROPIC_API_KEY` (matching original behavior — OAuth not used during init)

PR #342, opened 2026-03-17
fix: validate --temp-dir contents before smart_merge

voyax:
## Summary

- `zylos upgrade --temp-dir <dir>` only validated that the directory exists, not that it contains valid package contents
- If the dir exists but is empty (e.g. a temp dir from a previous `--check` that was already cleaned up), `smart_merge` treats the empty source as "new version has no files" and **deletes all deployed skill files**
- This caused Howard's `telegram 0.2.4` upgrade to wipe `src/bot.js` and `package.json`, breaking the service

## Fix

Add a `package.json` existence check on the provided `--temp-dir` for both the component upgrade and self-upgrade paths. Empty/invalid temp dirs now fail fast with a clear error message instead of silently wiping the installation.

## Test plan
- [ ] `zylos upgrade telegram --temp-dir /tmp/empty-dir` → should error: "missing package.json (empty or invalid)"
- [ ] `zylos upgrade telegram --temp-dir /tmp/valid-dir` (with package.json) → should proceed normally
- [ ] `zylos upgrade` (no --temp-dir) → unchanged behavior

### DIFF
--- a/cli/commands/component.js
+++ b/cli/commands/component.js
@@ -342,6 +342,31 @@ export async function upgradeComponent(args) {
 }
 
 /**
+ * Validate that a provided --temp-dir exists and contains a valid package.
+ * Prints/logs the error if invalid. Returns { valid: boolean }.
+ */
+function validateTempDir(tempDir, { jsonOutput, action, component }) {
+  const checks = [
+    { test: () => fs.existsSync(tempDir), msg: `Provided temp dir does not exist: ${tempDir}` },
+    { test: () => fs.existsSync(path.join(tempDir, 'package.json')), msg: `Provided temp dir is missing package.json (empty or invalid): ${tempDir}` },
+  ];
+  for (const { test, msg } of checks) {
+    if (!test()) {
+      if (jsonOutput) {
+        const errOutput = { action, success: false, error: msg };
+        if (component) errOutput.component = component;
+        errOutput.reply = formatC4Reply('error', { message: msg });
+        console.log(JSON.stringify(errOutput, null, 2));
+      } else {
+        console.error(`Error: ${msg}`);
+      }
+      return { valid: false };
+    }
+  }
+  return { valid: true };
+}
+
+/**
  * Handle --check flag: check for updates only (no lock needed).
  * Also fetches changelog, detects local changes, and runs Claude eval for a complete preview.
  * When --branch is specified, downloads from branch and reads version from its package.json.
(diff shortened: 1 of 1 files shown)


voyax (commented):
## Review by voya-nexus

### 1. DRY: Extract `validateTempDir()` helper

The temp-dir validation logic is duplicated across `handleUpgradeFlow` (line ~543) and `upgradeSelfCore` (line ~1005) — identical 22 lines. Extract into a shared helper:

```js
/**
 * Validate that a provided --temp-dir contains a valid package.
 * @param {string} tempDir
 * @param {boolean} jsonOutput
 * @param {'upgrade'|'self_upgrade'} action
 * @param {string} [component]
 * @returns {{ valid: boolean }} — if invalid, error is already printed/logged
 */
function validateTempDir(tempDir, { jsonOutput, action, component }) {
  const checks = [
    { test: () => fs.existsSync(tempDir), msg: `Provided temp dir does not exist: ${tempDir}` },
    { test: () => fs.existsSync(path.join(tempDir, 'package.json')), msg: `Provided temp dir is missing package.json (empty or invalid): ${tempDir}` },
  ];
  for (const { test, msg } of checks) {
    if (!test()) {
      if (jsonOutput) {
        const errOutput = { action, success: false, error: msg };
        if (component) errOutput.component = component;
        errOutput.reply = formatC4Reply('error', { message: msg });
        console.log(JSON.stringify(errOutput, null, 2));
      } else {
        console.error(`Error: ${msg}`);
      }
      return { valid: false };
    }
  }
  return { valid: true };
}
```

Then both call sites become one-liners:

```js
if (tempDirWasProvided) {
  if (!validateTempDir(tempDir, { jsonOutput, action: 'upgrade', component }).valid) return false;
  // ...
}
```

This also makes it trivial to add future checks (e.g. validate version field in package.json) without touching two places.

### 2. `checkAuth()` Stage 2 — timeout/network error should be `ok: true`, not `ok: false`

In `_httpValidateCredentials`, both `AbortError` (5s timeout) and network errors return `ok: false`. This is a behavioral regression:

- **Before:** Stage 2 (ping) treated `err.killed` (timeout), `ETIMEDOUT`, `ECONNREFUSED`, `ENOTFOUND` as transient → `ok: true` (uncertain)
- **After:** new Stage 2 (HTTP) treats the same conditions as auth failure → `ok: false`

In firewalled/Docker environments where `api.anthropic.com` is unreachable, this causes the activity monitor to incorrectly conclude auth has failed and kill the Claude process.

**Fix:** Only 401 should return `ok: false`. Everything else should be uncertain:

```js
async _httpValidateCredentials(authHeaders) {
  const controller = new AbortController();
  const timeoutId = setTimeout(() => controller.abort(), 5000);
  try {
    const resp = await fetch('https://api.anthropic.com/v1/models', {
      method: 'GET',
      headers: { 'anthropic-version': '2023-06-01', ...authHeaders },
      signal: controller.signal,
    });
    if (resp.status === 200) return { ok: true, reason: 'http_validate_200' };
    if (resp.status === 401) return { ok: false, reason: 'http_validate_401' };
    // 429, 5xx, other — don't block restart
    return { ok: true, reason: `http_validate_${resp.status}` };
  } catch {
    // Timeout or network error — can't verify, don't block restart
    return { ok: true, reason: 'http_validate_uncertain' };
  } finally {
    clearTimeout(timeoutId);
  }
}
```

This preserves the original conservative behavior: only a definitive 401 means "auth failed".

PR #496, opened 2026-04-14
fix(comm-bridge): self-heal better-sqlite3 on startup

zylos-luna:
## Summary
- replace hard import of `better-sqlite3` in `skills/comm-bridge/scripts/c4-db.js` with a guarded loader
- when load fails, run one-time startup repair (`npm install --omit=dev --ignore-scripts=false` + `npm rebuild better-sqlite3 --build-from-source`) and retry
- add explicit operator guidance in error message, including `ignore-scripts` fix and Node 22 LTS recommendation for very new Node majors
- document manual recovery steps in both `README.md` and `README.zh-CN.md`

## Validation
- `node --test skills/comm-bridge/scripts/__tests__/c4-db-control.test.js skills/comm-bridge/scripts/__tests__/c4-db-checkpoint.test.js skills/comm-bridge/scripts/__tests__/c4-db-cli.test.js`
- result: 66/66 passed

## Notes
- this addresses startup/install-time native-addon failures observed on Mac (darwin arm64) when npm scripts were skipped
- final confidence still needs environment verification on the reporter machine

### DIFF
--- a/README.md
+++ b/README.md
@@ -351,6 +351,24 @@ zylos list                    # List installed components
 zylos search [keyword]        # Search component registry
 ```
 
+### Native addon repair (`better-sqlite3`)
+
+If `comm-bridge` fails to start with a `better-sqlite3` binding error, Zylos now auto-attempts a one-time `npm` repair on startup. If it still fails, run:
+
+```bash
+npm config set ignore-scripts false
+cd ~/zylos/.claude/skills/comm-bridge
+npm install --omit=dev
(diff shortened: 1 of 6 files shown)


zylos-luna:
Closing: the root cause has been identified and fixed in PR #653 (process.execPath pinning). This PR's rebuild approach treated the symptom (ERR_DLOPEN) rather than the true cause (tmux clean PATH dropping nvm when PM2 strips it from the parent). #653 fixes the node resolution at the source.

PR #507, opened 2026-04-20
Fix settings hook migration for existing matcher groups

athanbase:
## Summary
- Fix settings hook sync so existing matcher groups receive newly added Zylos template hooks instead of only updating hooks already present.
- Preserve unrelated user hooks while appending missing core hooks by normalized script path.
- Align the self-upgrade fallback migration-hints path with matcher-aware sync semantics and prevent stale missing-hook hints from creating duplicates.

## Verification
- Reproduced the v0.4.13 -> current template scenario before the fix: `PostToolUseFailure` was added, but `session-foreground.js` stayed missing from `SessionStart[startup|clear|compact]`.
- Verified after the fix that `session-foreground.js` is added to all three SessionStart matchers and user hooks are preserved.
- `npm test`

### DIFF
--- a/cli/lib/__tests__/self-upgrade.test.js
+++ b/cli/lib/__tests__/self-upgrade.test.js
@@ -282,3 +282,138 @@ describe('Boolean setting migration hints (autoMemoryEnabled, autoDreamEnabled)'
     fs.rmSync(tmpDir, { recursive: true, force: true });
   });
 });
+
+function makeHook(scriptPath, timeout = 10000) {
+  return { type: 'command', command: `node ${scriptPath}`, timeout };
+}
+
+function makeMatcherGroup(matcher, scriptPaths) {
+  return {
+    matcher,
+    hooks: scriptPaths.map(p => makeHook(p)),
+  };
+}
+
+describe('settings hook migration hints', () => {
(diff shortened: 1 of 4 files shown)


zylos01 (approved):
Re-read the matcher-aware hook migration changes and spot-checked the self-upgrade + sync path. No blocking issues found.

jinglever (approved):
No blocking issue.

The matcher-aware migration change looks correct. `generateMigrationHints` now compares template hooks only against the corresponding matcher group, so missing hooks inside an already-existing group are detected instead of being masked by another matcher. `applyMigrationHints` also applies missing/modified/removed hook hints to the hinted matcher group, and the stale missing-hook guard avoids duplicating hooks that are already present at apply time. `syncHooks` now backfills missing template hooks into existing matcher groups while preserving user hooks, which matches the intended migration behavior for historical settings.

Validation I ran:
- `node --check cli/lib/self-upgrade.js` passed.
- `node --check cli/lib/sync-settings-hooks.js` passed.
- After `npm ci`, targeted node tests passed: `cli/lib/__tests__/self-upgrade.test.js` and `cli/lib/__tests__/sync-settings-hooks.test.js` (47/47).
- `gh pr checks` reports no checks on this branch.

PR #509, opened 2026-04-21
fix(upgrade): back up preserve-declared dataDir files before upgrade

### LINKED ISSUE #502: Upgrade backup should include dataDir (database files not backed up)
## Problem

The 7-step upgrade pipeline (`cli/lib/upgrade.js`) creates a backup in Step 2 (`step2_backup`), but only backs up `skillDir` (`~/.claude/skills/<component>/`). The `dataDir` (`~/zylos/components/<component>/`) is **not included** in the backup.

This means database files (e.g., `recruit.db`), config files, and other runtime data in the component's data directory are not preserved before an upgrade.

### Impact

During a Recruit ATS upgrade (v0.2.7 → v0.2.8), the migration script contained a bug that triggered `CASCADE DELETE` on the production SQLite database, destroying all candidate, role, evaluation, and interview data. Because `dataDir` was not backed up, the data was unrecoverable.

- 35 candidates lost
- 108 evaluations lost
- 28 internal interviews + 160 messages lost
- No backup available for rollback

### Root Cause

In `upgrade.js` line 387-399, `step2_backup` only copies `ctx.skillDir`:

```javascript
function step2_backup(ctx) {
  const timestamp = new Date().toISOString().replace(/[:.]/g, '-');
  const backupDir = path.join(ctx.skillDir, '.backup', timestamp);
  copyTree(ctx.skillDir, backupDir, { excludes: ['node_modules', '.backup', '.zylos'] });
  // ...
}
```

`ctx.dataDir` is defined in `createContext()` (line 321) but never used in the backup step.

## Proposed Fix

Add `dataDir` backup to `step2_backup`:

1. Copy `dataDir` contents (especially `.db`, `.json`, `.sqlite` files) to the backup directory alongside the skill backup
2. Include `dataDir` in the rollback path so failed upgrades can restore data
3. Consider backing up only data files (not logs) to keep backup size reasonable

## Additional Context

- `dataDir` is already referenced in uninstall and info commands (`component.js`)
- The cleanup logic (`cleanOldBackups`) already retains only the latest backup, so storage impact is minimal
- Components with SQLite databases (recruit, comm-bridge, scheduler) are especially vulnerable without data backup

voyax:
## Summary

- `step2_backup` now reads the `preserve` list from SKILL.md and backs up files with the `data://` prefix from `dataDir` into `_datadir/` within the backup directory
- `rollback` restores `dataDir` files from `_datadir/` when an upgrade fails, with `mkdirSync` to handle the directory-deleted edge case
- `syncTree` in rollback excludes `_datadir` to avoid mixing data files into skillDir

## Design

Uses the existing `preserve` field in SKILL.md — no structural change. Components declare critical dataDir files with a `data://` prefix:

```yaml
preserve:
  - config.json           # skillDir (no change)
  - data://recruit.db     # dataDir (new)
```

`preserve` is backup-only, not a write lock. On successful upgrade, all files are updated normally (skillDir by smartSync, dataDir by migration). The backup is only used on rollback.

See #502 for full design discussion.

## Test plan

- [ ] Upgrade a component with `data://` entries in preserve — verify `_datadir/` appears in `.backup/<timestamp>/` with declared files
- [ ] Force a failed upgrade — verify rollback restores dataDir files from `_datadir/`
- [ ] Upgrade a component without `data://` entries — verify no `_datadir/` created, existing behavior unchanged
- [ ] Upgrade a component with no preserve field — verify no error, backup completes normally
- [ ] Verify `data://` entry for non-existent file is skipped without error

Closes #502

### DIFF
--- a/cli/lib/upgrade.js
+++ b/cli/lib/upgrade.js
@@ -425,7 +425,8 @@ function step2_backup(ctx) {
         const srcPath = path.resolve(ctx.dataDir, entry);
         // Path traversal guard: resolved path must stay within dataDir
         if (!srcPath.startsWith(resolvedDataDir + path.sep)) continue;
-        if (fs.existsSync(srcPath) && fs.statSync(srcPath).isFile()) {
+        // Use lstatSync to reject symlinks (prevents symlink-based traversal escape)
+        if (fs.existsSync(srcPath) && fs.lstatSync(srcPath).isFile()) {
           const destPath = path.join(dataBackupDir, entry);
           fs.mkdirSync(path.dirname(destPath), { recursive: true });
           fs.copyFileSync(srcPath, destPath);
--- a/cli/lib/upgrade.js
+++ b/cli/lib/upgrade.js
@@ -383,6 +383,7 @@ function step1_stopService(ctx) {
 
 /**
  * Step 2: filesystem backup to .backup/<timestamp>/
+ * Backs up skillDir (source code) and preserve-declared dataDir files (runtime data).
  */
 function step2_backup(ctx) {
   const startTime = Date.now();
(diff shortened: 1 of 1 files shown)


jinglever (approved):
No blocking issue.

This is the safer direction for the dataDir rollback gap: it backs up only `data://` entries declared in SKILL.md preserve metadata, reads preserve entries from both old `skillDir` and new `tempDir` so the first release introducing `data://` entries is protected, keeps resolved paths within dataDir, creates parent directories for nested entries, and rejects symlinks with `lstatSync(...).isFile()` before copying. Rollback restores `_datadir` recursively, so nested preserve files are covered.

Validation I ran:
- `node --check cli/lib/upgrade.js` passed.
- After `npm ci`, targeted node tests passed: `cli/lib/__tests__/upgrade-restart.test.js` and `cli/lib/__tests__/fs-utils.test.js` (6/6).
- `gh pr checks` reports no checks on this branch.

Non-blocking follow-up: this path would benefit from direct unit coverage for `data://` preserve backup/rollback cases, especially nested files, path traversal attempts, and symlink rejection. The current implementation handles those cases in code, but the coverage would prevent regressions in this safety-critical path.

PR #510, opened 2026-04-21
fix(init): accept non-sk-ant- API keys for custom base URLs

danielzhou82:
`validateInitOptions` and the interactive API-key prompt currently hard-reject any `--api-key` / `ANTHROPIC_API_KEY` that does not start with `sk-ant-`. That check is safe for `api.anthropic.com`, but it breaks deployments that point `ANTHROPIC_BASE_URL` at a proxy, gateway, or key-translation layer whose native keys use a different prefix (e.g. `cocoak-*` from sub2api's COCOAK middleware, or arbitrary strings from third-party relays). In those environments zylos init exits before reaching the Claude Code install step, which then cascades into a stuck / login state at runtime.

Introduce a small `isCustomAnthropicEndpoint(baseUrl)` helper and skip the positive `sk-ant-` prefix check when the endpoint resolves to any host other than `api.anthropic.com` (or a subdomain thereof). The "this looks like a setup token, not an API key" guard for `sk-ant-oat-*` is preserved for both paths, since that mistake is common and orthogonal to the endpoint.

Tests cover: official vs custom host detection (including upper-case hosts, `*.anthropic.com` subdomains, HTTP URLs, and malformed strings); that `cocoak-*` is rejected against the default endpoint and accepted against a custom one; that `sk-ant-oat-*` is still rejected as an API key on custom endpoints; and that `sk-ant-*` still passes the default path.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Checklist

- [ ] Code follows the project's ESM-only style
- [ ] Self-review completed
- [ ] Changes are tested locally
- [ ] CHANGELOG.md updated (if applicable)
- [ ] No secrets or internal references included


voyax (dismissed):
Code review: **Approve** ✅

Clean, well-scoped fix. Reviewed `isCustomAnthropicEndpoint`, both validation and interactive prompt paths, and the test suite.

**What's correct:**
- `isCustomAnthropicEndpoint` properly treats all `*.anthropic.com` subdomains as official (covers beta/staging endpoints)
- Setup token rejection (`sk-ant-oat-*`) is preserved for all endpoints — correct, since that mistake is prefix-agnostic
- Malformed URLs in `isCustomAnthropicEndpoint` fall through to `true` (custom), but `isValidBaseUrl` upstream already catches those, so no real risk
- Error message now hints at `--base-url` when a non-standard key is rejected against the default endpoint — good UX
- Interactive prompt text adapts to custom endpoints
- Tests are thorough: 5 new test cases covering official/custom detection, acceptance, rejection, and the setup-token guard

**No issues found.** The ordering of checks is correct (prefix check → setup token check), the helper is properly exported for testability, and the change is backward-compatible for the default `api.anthropic.com` path.

leslielin2025:
Problem is real, but the fix is still incomplete.

Relaxing the `sk-ant-` prefix check for custom `ANTHROPIC_BASE_URL` is the right direction, because the current init flow does incorrectly reject valid proxy/gateway-issued keys up front.

However, both the non-interactive and interactive paths still call `verifyApiKey()`, and that verifier is hardcoded to `https://api.anthropic.com/v1/messages`. So for custom endpoints, this PR changes the failure mode from “prefix rejected” to “verification against the wrong endpoint failed”.

In other words, custom-key init is still blocked end-to-end.

I think this should either:
- pass `baseUrl` through to `verifyApiKey()` and validate against the configured endpoint, or
- skip online verification for custom endpoints and only persist the key with a clear runtime-validation warning.

voyax (approved):
Re-review after v2 rewrite: **Approve** ✅

Significantly expanded scope from v1 — now includes `verifyApiKey` routing through custom endpoints. Review covers all new code.

**New additions reviewed:**

1. **`resolveVerifyEndpoint(baseUrl)`** — Correctly decomposes base URL into hostname/port/protocol/path. Handles path prefixes (e.g., `https://proxy/anthropic` → `/anthropic/v1/messages`). Falls back to `api.anthropic.com` for null, empty, malformed, or non-http/https URLs.

2. **`verifyApiKey(apiKey, baseUrl)`** — Clean dual-behavior design:
   - Official endpoint: strict, fail-closed (network errors → invalid). Preserves prior behavior.
   - Custom endpoint: lenient, only 401 → invalid. Network errors/timeouts → saved-but-unverified with warning. This is the right tradeoff — third-party proxies that don't implement the empty-POST probe shouldn't block init.

3. **Return type change** (`boolean` → `{ valid, authError?, unverified?, reason? }`) — Both call sites (non-interactive + interactive) handle all variants correctly. UX messages differentiate auth errors vs transport failures vs unverified-but-proceed.

4. **`http` import added** — needed for `http://` custom endpoints. `requester` selection via protocol is clean.

5. **Tests** — 17 tests, all passing. `verifyApiKey` tests use a real stub HTTP server (startStubServer pattern) — tests 200/400/401 responses and network errors. Good coverage.

**No issues found.** The `lenientResult` helper keeps the dual-behavior logic DRY. CHANGELOG accurately describes both changes. Test file wired into `test:node` script for CI.

zylos01:
## MEDIUM #2: Interactive path missing `sk-ant-oat` rejection

**Head reviewed**: `9cf4113d`

The `--api-key` flag path (`validateInitOptions`, line 1799) correctly rejects `sk-ant-oat-*` with a clear message: *"That looks like a setup token, not an API key."*

The interactive prompt path (Option 2 — "Anthropic API key", line 2283) does not have this check. The only validation there is:

```js
if (!apiKey.startsWith('sk-ant-') && !customEndpoint) {
  // "Invalid format. API key should start with sk-ant-"
}
```

Since `sk-ant-oat-*` starts with `sk-ant-`, it passes through. The token then goes to `verifyApiKey()`:
- **Official endpoint**: returns 401 → user sees "API key is invalid" (no hint that they pasted a setup token in the wrong slot)
- **Custom endpoint** (the scenario this PR targets): lenient mode may save-but-unverified → setup token silently stored as an API key

The commit message states *"Setup tokens (`sk-ant-oat-*`) are still rejected in the `--api-key` slot regardless of endpoint"* — true for the flag path, not for the interactive path.

**Suggested fix** (at line ~2282, before the existing `startsWith('sk-ant-')` check):

```js
} else if (apiKey.startsWith('sk-ant-oat')) {
  console.log(`  ${error('That looks like a setup token, not an API key.')}`);
  console.log(`    ${dim('Cancel and choose option 3 (Setup token) instead.')}`);
} else if (!apiKey.startsWith('sk-ant-') && !customEndpoint) {
  console.log(`  ${error('Invalid format. API key should start with sk-ant-')}`);
```

~5 lines, symmetric with the flag path.

— zylos01 (review layer: intent/scope/design)

jinglever:
MEDIUM: `resolveVerifyEndpoint()` double-appends `/v1` when users pass the documented base URL shape.

`validateInitOptions()` tells users to run `zylos init --base-url https://example.com/v1`, but `resolveVerifyEndpoint('https://example.com/v1')` currently returns `/v1/v1/messages`. That means custom endpoint verification probes the wrong path for the common OpenAI/Anthropic-compatible base URL convention where the configured base URL already includes `/v1`.

Evidence:
- `cli/commands/init.js:424-430` keeps the full pathname and appends `/v1/messages` unconditionally.
- `cli/commands/init.js:1813-1815` documents `https://example.com/v1` as the expected input shape.
- Existing resolver tests cover root paths and `/anthropic`, but not a base URL already ending in `/v1`.

Local reproduction on head `9cf4113d`:

```text
https://example.com/v1 => /v1/v1/messages
https://example.com/anthropic => /anthropic/v1/messages
https://example.com => /v1/messages
```

Expected behavior should either normalize a trailing `/v1` before appending `/messages`, or document that Anthropic `--base-url` must be the origin/prefix before `/v1` and reject/adjust the current example.

PR #521, opened 2026-05-04
fix(caddy): forward stripped route prefix

jinglever:
## Summary
- generate `X-Forwarded-Prefix` automatically for component Caddy routes that declare `strip_prefix`
- render stripped reverse proxy routes as a `reverse_proxy { ... }` block with `header_up X-Forwarded-Prefix <strip_prefix>`
- keep non-stripped reverse proxy routes as the existing single-line directive

## Why
Components like `recruit` are mounted externally at `/recruit/*` while the upstream app receives stripped root paths. Sending `X-Forwarded-Prefix` lets components infer the browser-visible base path without hardcoding their external mount prefix.

## Verification
- `node --check cli/lib/caddy.js`
- `node --experimental-vm-modules node_modules/.bin/jest test/caddy.test.js`
- `git diff --check`

### DIFF
--- a/cli/lib/caddy.js
+++ b/cli/lib/caddy.js
@@ -36,7 +36,7 @@ export function isCaddyAvailable() {
  * @param {Array} httpRoutes - Array of { path, type, target, strip_prefix? }
  * @returns {string} Caddy configuration block (indented for domain block)
  */
-function generateRouteBlocks(httpRoutes) {
+export function generateRouteBlocks(httpRoutes) {
   const lines = [];
   for (const route of httpRoutes) {
     if (route.type === 'reverse_proxy') {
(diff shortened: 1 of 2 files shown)


zylos01 (approved):
## Review: APPROVE ✅

Small, clean change. Reviewed `caddy.js` and the new tests.

### What it does
When a component route has `strip_prefix` (e.g., recruit mounted at `/recruit/*`), the generated Caddyfile now sends `X-Forwarded-Prefix` to the upstream, so the app can discover its external mount path without hardcoding it.

### Caddy syntax check
Generated output for a stripped route:
```
    handle /recruit/* {
        uri strip_prefix /recruit
        reverse_proxy localhost:3465 {
            header_up X-Forwarded-Prefix /recruit
        }
    }
```
Valid Caddyfile syntax. `header_up` is a standard `reverse_proxy` subdirective. Indentation consistent with the existing structure.

### Code review
- `strip_prefix` is checked twice — once for `uri strip_prefix` (modify URI before proxying) and once for the `reverse_proxy` block form (add header). Two separate concerns, correct to separate. Could consolidate into one `if` block with both operations, but current form is clear enough — non-blocking.
- Non-stripped routes unchanged (single-line `reverse_proxy` directive). Backward compatible.
- `generateRouteBlocks` exported for testability — fine, it's a pure utility function.

### Header choice
`X-Forwarded-Prefix` is de facto standard (Spring Boot, Django, Rails, etc.). Good choice. This directly supports components like recruit that need to know their external base path — complements PR #34's auth route fix.

### Tests
Two tests covering both branches (stripped → block form with header, non-stripped → single-line without header). Assertions check both presence and absence of the header. Clean.

### No concerns
- No security implications (header only goes upstream, not to the client)
- Additive change — existing apps that don't read the header are unaffected

LGTM.

zz-howard (approved):

zz-howard (approved):

PR #530, opened 2026-05-06
fix(comm-bridge): treat "-" as stdin marker in c4-send 3-arg form

jessie-coco:
## Summary

Before: `node c4-send.js <channel> <endpoint> -` with piped stdin would send the literal string `-` as the message, because the 3-arg branch in `main()` unconditionally read `cleanArgs[2]` as the message body and ignored stdin.

This silently broke heredoc-composed long-form messages:

```bash
node c4-send.js lark <chat_id> - <<'EOF'
<multi-line content>
EOF
```

The intent (read message from stdin) never fired — recipients on Lark / HxA Connect saw a literal `-` for what should have been a multi-paragraph briefing.

The pattern is documented elsewhere in the codebase: `lark-cli api --data -`, `--params -`, `--markdown -`, `c4-control.js` etc. all use `-` to mean "read from stdin". `c4-send.js` was the outlier.

## Fix

In `c4-send.js`, when 3 args are passed AND `cleanArgs[2] === '-'` AND stdin is available, read the message from stdin via the existing `readStdin()` helper.

The literal-`-` path remains for the no-stdin case (backward compatibility), so existing callers passing `-` as a literal message without piping still work.

## Test plan

- [x] New `c4-send stdin "-" marker` describe block in `c4-send-cli.test.js` with 2 cases:
  - 3 args + `-` + piped stdin → reads stdin (previously sent literal `-`)
  - Multi-line stdin content (heredoc-style, newlines preserved end-to-end)
- [x] Locally verified: 2 new tests pass, no regression in the 5 other tests that pass on main
- [x] 3 pre-existing test failures on main (`broadcast send`, `channel script not found`, `failed channel reports failure`) are **unrelated to this change** — they fail on main as well

## Real-world impact (already observed)

Before this fix, several real conversations on 2026-05-06 lost content:
- Two competitive-research briefings (VNTR + Sierra) on a Lark group
- One MR review request via HxA Connect DM
- Several long status updates on Lark DM

All recipients saw a literal `-` in place of the message body. Re-sends after the local hotfix went through correctly. Upstreaming so other agents using the same heredoc pattern don't keep dropping content.

### DIFF
--- a/skills/comm-bridge/scripts/__tests__/c4-send-cli.test.js
+++ b/skills/comm-bridge/scripts/__tests__/c4-send-cli.test.js
@@ -8,11 +8,12 @@ import { fileURLToPath } from 'node:url';
 
 const CLI_PATH = fileURLToPath(new URL('../c4-send.js', import.meta.url));
 
-function cli(args, env = {}) {
+function cli(args, env = {}, stdinInput = undefined) {
   const result = spawnSync('node', [CLI_PATH, ...args], {
     env: { ...process.env, ...env },
     encoding: 'utf8',
-    timeout: 5000
+    timeout: 5000,
+    input: stdinInput
   });
   return { stdout: result.stdout, stderr: result.stderr, status: result.status };
 }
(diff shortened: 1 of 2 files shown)


zylos01:
Decision: keep this PR pending for now.

The direction we want is broader than this targeted `-` stdin-marker fix: move `c4-send.js` to a stdin-only message contract, where argv carries only routing (`channel`, `endpoint`) and message body always comes from stdin. Under that contract, old three-argument message calls such as:

```bash
node c4-send.js lark "endpoint" "message"
```

should fail with a clear error and usage hint instead of being sent.

However, the AM refactor PR is still in flight and may contain many existing `c4-send.js` call sites using the old form. To avoid churn and conflicts, wait until the AM refactor lands in `main`, then do the migration and contract change in one dedicated branch:

- migrate all first-party/internal `c4-send.js` calls to stdin;
- update docs, skill instructions, and prompt examples to stdin-only;
- change `c4-send.js` so three-argument message calls hard-fail with a helpful error showing the correct stdin/heredoc usage;
- add tests for stdin success and three-argument failure.

Labeling this PR as `pending` until that follow-up branch is ready.

PR #543, opened 2026-05-06
fix: guard against duplicate memory-sync subagents in new-session

zylos01:
## Summary

- Adds step 1 to the new-session pre-switch checklist: check `TaskList` for running memory-sync subagents before triggering a new one
- If a memory-sync is already running, skip spawning another — prevents the deadlock where repeated context-monitor triggers create contending subagents
- Re-numbered existing steps (1→2, 2→3, 3→4, 4→5) and fixed internal cross-references

## Context

Forensic analysis of a btwlei instance deadlock showed: context-monitor fires `new-session` every 6 min when context stays above threshold, but each trigger spawns a new memory-sync subagent while the previous one is still running. Multiple concurrent syncs + user message handling caused the main loop to hang for 35 min until Claude's internal watchdog killed the process.

Howard's direction: check for running memory-sync subagents in the new-session prompt, don't add ack/debounce mechanisms.

## Test plan

- [ ] Verify new-session skill triggers correctly when no memory-sync is running
- [ ] Verify duplicate memory-sync is skipped when one is already in progress
- [ ] Confirm step numbering and cross-references are consistent

### DIFF
--- a/skills/new-session/SKILL.md
+++ b/skills/new-session/SKILL.md
@@ -20,7 +20,14 @@ Use runtime-specific switch commands:
 
 Before sending the session switch command, complete these steps **in order**:
 
-### 1. Inventory running background tasks
+### 1. Check for running memory-sync subagents
+
+Before doing anything else, check if a memory-sync subagent is already running (use `TaskList` on Claude, or the equivalent on other runtimes). Look for any task whose description contains "memory sync" or "zylos-memory".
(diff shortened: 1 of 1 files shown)


zylos01:
Wrong location — memory sync is no longer triggered by new-session. Moving the guard to zylos-memory skill instead.

PR #570, opened 2026-05-08
fix(upgrade): verify activity-monitor env via pm2 jlist

leslielin2025:
## Summary
- Replace self-upgrade step 11 activity-monitor env verification from `pm2 env "activity-monitor"` to `pm2 jlist` parsing.
- Avoid PM2 v6 behavior where `pm2 env` accepts numeric `pm_id` but rejects process names.
- Add coverage that the default verifier uses `pm2 jlist`, does not call `pm2 env activity-monitor`, and still fails before `pm2 save` when `ZYLOS_PACKAGE_ROOT` is missing.

## Validation
- `node --test cli/lib/__tests__/upgrade-restart.test.js`
- `node --test cli/lib/__tests__/pm2.test.js`
- `npm test`

### DIFF
--- a/cli/lib/__tests__/upgrade-restart.test.js
+++ b/cli/lib/__tests__/upgrade-restart.test.js
@@ -86,7 +86,7 @@ describe('step11_startCoreServices', () => {
 
     fs.mkdirSync(binDir, { recursive: true });
     fs.writeFileSync(ecosystemPath, 'module.exports = { apps: [] };\n', 'utf8');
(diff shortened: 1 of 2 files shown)


zylos01 (approved):
Codex R1: CLEAN

Reviewed the full PR diff and the surrounding `step11_startCoreServices` restart/verify flow. The switch from `pm2 env activity-monitor` to `pm2 jlist` avoids the PM2 v6 name lookup failure while still verifying the refreshed `ZYLOS_PACKAGE_ROOT` before `pm2 save`.

Also checked local `pm2 jlist` shape: `activity-monitor.pm2_env.ZYLOS_PACKAGE_ROOT` is present as expected.

Validation run locally:
- `node --test cli/lib/__tests__/upgrade-restart.test.js`
- `npm test`

PR #572, opened 2026-05-08
fix(upgrade): run self-upgrade post-install steps from new package

leslielin2025:
## Summary
- after `npm install -g` completes, hand off self-upgrade steps 5-12 to the newly installed package's `self-upgrade-finalize.js`
- preserve the existing `--branch` behavior: install the latest code from the branch at download time
- do not rollback after the finalizer handoff; post-install failures are reported from the new package path

## Validation
- node --test cli/lib/__tests__/self-upgrade.test.js cli/lib/__tests__/upgrade-restart.test.js
- npm test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to zylos-core will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Upgrade Notes
+- If self-upgrade fails with `failed to verify activity-monitor PM2 env after restart`, first run `pm2 stop activity-monitor`, then retry with `zylos upgrade --self -y`.
+
 ## [0.4.13] - 2026-04-12
 
 ### Fixed
--- a/cli/lib/__tests__/self-upgrade.test.js
+++ b/cli/lib/__tests__/self-upgrade.test.js
@@ -18,7 +18,6 @@ describe('self-upgrade finalizer handoff', () => {
     assert.deepEqual(createFinalizeState({
       tempDir: '/tmp/new-core',
       backupDir: '/tmp/backup',
-      servicesStopped: ['activity-monitor'],
       servicesWereRunning: ['activity-monitor', 'c4-dispatcher'],
       from: '0.4.12',
       to: '0.4.13',
@@ -28,7 +27,6 @@ describe('self-upgrade finalizer handoff', () => {
       schemaVersion: 1,
       tempDir: '/tmp/new-core',
       backupDir: '/tmp/backup',
-      servicesStopped: ['activity-monitor'],
       servicesWereRunning: ['activity-monitor', 'c4-dispatcher'],
       from: '0.4.12',
       to: '0.4.13',
@@ -75,8 +73,7 @@ describe('self-upgrade finalizer handoff', () => {
     }]);
   });
 
-  it('rolls back with restored state when a post-install step fails', () => {
-    const rollbackCalls = [];
+  it('fails without rollback when a post-install step fails', () => {
     const result = runSelfUpgradeFinalize({
       schemaVersion: 1,
       tempDir: '/tmp/new-core',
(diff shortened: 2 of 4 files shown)


zylos01 (changes requested):
Codex R1: request changes.

Findings:

1. P1: Finalizer/post-install failures can leave core services stopped.

`runSelfUpgrade` stops previously running core services in step 3, then installs the new global package in step 4 and hands off to the installed finalizer. In the new failure path, `runSelfUpgradeFinalize` returns `buildSelfUpgradeResult(ctx, failedStep, null, false)` and `runSelfUpgrade` also returns with `rollback.performed=false` when the installed finalizer throws. There is no recovery path that restarts `ctx.servicesWereRunning`.

If any new-code finalizer step before service restart fails, for example sync skills, npm dependency install, template sync, hook install, or a finalizer crash, the instance can be left with `activity-monitor`, C4 dispatcher, and other core services stopped. Avoiding rollback after the package handoff is reasonable, but the failure path still needs to restore liveness from the new package, or at least restart the services that step 3 stopped before returning failure.

Relevant code:
- `cli/lib/self-upgrade.js:1531`
- `cli/lib/self-upgrade.js:1596`

Suggested fix: add a new-package recovery/finally path for finalizer failures that restarts previously running services, or move service restart into a guaranteed cleanup path after step 4. The important invariant is that post-install failure must not leave the agent offline.

2. P2: Branch self-upgrade is still downloading a moving branch, not a resolved commit.

The PR title/body says branch targets are resolved to a concrete commit SHA before download, but the implementation still calls `downloadBranch(REPO, branch, tempDir)` and the component flow still passes only the branch name into `downloadCoreToTemp`. Logs/results only expose `branch:<name>` and do not carry requested ref/resolved commit metadata.

That means `zylos upgrade --self --branch main` can still drift if `main` moves between check/changelog/confirmation and the actual archive download. It also leaves no installed commit hash for rollback/debugging, which was the core requirement discussed in the thread.

Relevant code:
- `cli/lib/self-upgrade.js:137`
- `cli/commands/component.js:1007`
- `cli/commands/component.js:1033`

Suggested fix: resolve the branch to a SHA once, download the archive for that SHA, and pass `requestedRef` / `resolvedCommit` through download result, upgrade result, logs, and any persisted metadata.

Verification run locally:
- `node --test cli/lib/__tests__/self-upgrade.test.js cli/lib/__tests__/upgrade-restart.test.js`
- `npm test`

Both passed, but the two issues above are behavioral gaps not covered by the current tests.

zylos01 (changes requested):
Codex R2: 0 P1 + 1 P2 + 0 P3

1. P2: `--branch` self-upgrade still installs a moving branch ref instead of a resolved commit.
   File: `cli/lib/self-upgrade.js:133`
   Issue: `checkForCoreUpdates({ branch })` reads `package.json` from the branch during the check phase, but the install path later calls `downloadCoreToTemp(check.latest, branch)`, and `downloadCoreToTemp()` calls `downloadBranch(REPO, branch, tempDir)`. That still fetches `refs/heads/<branch>` at download time, so if `main` or the PR branch moves between check/confirmation and download, the installed package may differ from the version/commit the user reviewed. The UI/result also still reports `branch:<name>` rather than the resolved SHA. This does not satisfy the PR title/contract of pinning branch targets to a commit.
   Fix: Resolve the branch to an immutable commit SHA before confirmation/install, carry that SHA through the check/download/result path, and download the tarball by that SHA (GitHub tarball API supports commit refs). Also surface the resolved SHA in JSON/human output and add a regression test that proves `--branch main` downloads the resolved SHA rather than the mutable branch name.

The new finalizer handoff looks aligned with linfan's clarification: post-install steps 5-12 are invoked from the newly installed package, and post-handoff failures no longer rollback. I did not keep the earlier rollback objection as a blocker.

Verification:
- `node --test cli/lib/__tests__/self-upgrade.test.js cli/lib/__tests__/upgrade-restart.test.js`
- `npm test`

zylos01 (approved):
Codex R3: CLEAN under the updated requirement.

Given the clarified product decision that `--branch` should keep installing the latest branch code and does not need a resolved SHA pin, I withdraw the previous branch-drift blocker. The remaining code path looks consistent with the revised scope: step 4 installs the new package, then steps 5-12 are delegated to `self-upgrade-finalize.js` from the newly installed package, and post-handoff failures intentionally do not rollback.

Non-blocking: the PR title/description still says "pin self-upgrade branch targets to commit". Please rename/update it so future reviewers do not evaluate the branch path against the old resolved-SHA contract.

Verification:
- `node --test cli/lib/__tests__/self-upgrade.test.js cli/lib/__tests__/upgrade-restart.test.js`
- `npm test`

PR #575, opened 2026-05-08
feat: ZYLOS_TMUX_ENV manifest for tmux session env injection

zylos01:
## Summary

- Adds `ZYLOS_TMUX_ENV` manifest variable support: users declare which `.env` variables should be injected into the Claude/Codex tmux session at launch time
- Both `ClaudeAdapter` and `CodexAdapter` read the manifest and inject resolved variables via sourced temp file (same secure pattern as auth tokens)
- Works for both new tmux sessions and existing session restarts

## Motivation

OTel telemetry and other Claude Code features require specific environment variables (`OTEL_*`, `CLAUDE_CODE_ENABLE_TELEMETRY`, etc.) to be present in the tmux session. Previously these had to be set in `~/.bashrc`, which:
1. Is a system-level file that zylos components shouldn't modify
2. Doesn't work for PM2-launched tmux sessions (non-interactive shell)

This PR introduces a declarative mechanism: the user lists variable names in `.env`, and zylos-core handles injection. No code changes needed to add/remove variables.

## Usage

```env
# Declare which variables to inject into the tmux session
ZYLOS_TMUX_ENV=OTEL_METRICS_EXPORTER,OTEL_EXPORTER_OTLP_ENDPOINT,CLAUDE_CODE_ENABLE_TELEMETRY

# Define the actual values
OTEL_METRICS_EXPORTER=otlp
OTEL_EXPORTER_OTLP_ENDPOINT=http://127.0.0.1:4318
CLAUDE_CODE_ENABLE_TELEMETRY=1
```

## Implementation

- `_readManifestEnvVars(envContent)` reads `ZYLOS_TMUX_ENV`, splits by comma, resolves each variable from `.env`
- Resolved vars are written to a temp file (mode 0600), sourced with `set -a`, then deleted — no exposure in `ps`/`/proc/cmdline`
- Claude adapter: manifest vars merged with existing auth token injection path
- Codex adapter: manifest vars injected via env prefix before codex command

## Test plan

- [ ] 11 unit tests covering: manifest parsing, variable resolution, missing vars, quoted values, spaces, empty manifest
- [ ] Verify on zylos01: add `ZYLOS_TMUX_ENV` to `.env`, restart, confirm env vars are present in Claude session
- [ ] Verify OTel data flows to OTLP receiver after restart

### DIFF
--- a/cli/lib/__tests__/tmux-env-manifest.test.js
+++ b/cli/lib/__tests__/tmux-env-manifest.test.js
@@ -1,0 +1,102 @@
+import assert from 'node:assert/strict';
+import { describe, test } from 'node:test';
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+
+const tmpRoot = fs.mkdtempSync(path.join(os.tmpdir(), 'zylos-tmux-env-'));
+process.env.HOME = tmpRoot;
+process.env.ZYLOS_DIR = path.join(tmpRoot, 'zylos');
+fs.mkdirSync(process.env.ZYLOS_DIR, { recursive: true });
+fs.writeFileSync(path.join(process.env.ZYLOS_DIR, '.env'), '', 'utf8');
+
+const { _parseEnvValue, _readManifestEnvVars } = await import('../runtime/claude.js');
+
+describe('ZYLOS_TMUX_ENV manifest', () => {
+  test('_readManifestEnvVars returns empty array when ZYLOS_TMUX_ENV is absent', () => {
+    const result = _readManifestEnvVars('FOO=bar\nBAZ=qux\n');
+    assert.deepEqual(result, []);
+  });
+
+  test('_readManifestEnvVars resolves declared variables from .env content', () => {
+    const env = [
+      'ZYLOS_TMUX_ENV=OTEL_ENDPOINT,CLAUDE_CODE_ENABLE_TELEMETRY,MY_VAR',
+      'OTEL_ENDPOINT=http://localhost:4318',
+      'CLAUDE_CODE_ENABLE_TELEMETRY=1',
+      'MY_VAR=hello world',
+    ].join('\n');
+
+    const result = _readManifestEnvVars(env);
+    assert.deepEqual(result, [
+      { key: 'OTEL_ENDPOINT', value: 'http://localhost:4318' },
+      { key: 'CLAUDE_CODE_ENABLE_TELEMETRY', value: '1' },
+      { key: 'MY_VAR', value: 'hello world' },
+    ]);
+  });
+
+  test('_readManifestEnvVars skips variables not defined in .env', () => {
+    const env = [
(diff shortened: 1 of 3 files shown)


jinglever (changes requested):
Found one blocking issue in the ZYLOS_TMUX_ENV path.\n\nThe manifest values are written into a shell-sourced temp file as raw `KEY='value'` lines, without validating `KEY` as a shell identifier and without shell-safe quoting `value`. This affects both runtime implementations:\n\n- `cli/lib/runtime/claude.js`: manifest names are accepted as-is in `_readManifestEnvVars()`, then emitted via `=''` into the sourced temp file.\n- `cli/lib/runtime/codex.js`: same parser/emitter pattern.\n\nA valid-looking .env value containing a single quote breaks the temp file, e.g. `QUOTE=it's fine` becomes `QUOTE='it's fine'`, so the runtime launch will fail or source a truncated/broken file. A malformed manifest name such as `BAD-NAME` is also accepted and emitted as `BAD-NAME='oops'`, which is not a valid shell assignment. Since this file is sourced before launching Claude/Codex, the parser should reject names that do not match shell env identifiers, e.g. `^[A-Za-z_][A-Za-z0-9_]*$`, and values should be emitted using a shell-safe quoting helper. The simplest safe single-quote form is `'` + value.replaceAll(`'`, `'\''`) + `'`, or avoid shell parsing entirely by passing env through tmux `-e` / child process env where possible.\n\nPlease add tests for both cases: declared variable with a single quote in the value, and invalid manifest variable names.\n\nOther requested review points looked OK at the design level: the new-session path sources before launch, the existing-session path follows the same restart-command pattern already used by the adapters, and Codex's prefix style is semantically equivalent to Claude's envParts style once the shared quoting/validation issue is fixed.\n\nLocal verification run on the PR branch: `node --test cli/lib/__tests__/tmux-env-manifest.test.js cli/lib/__tests__/codex.test.js cli/lib/__tests__/runtime-setup.test.js` passed: 23/23.

jinglever (approved):
Re-review after fix: the blocking shell env injection issue is resolved.\n\nVerified at head 1dced73:\n- Manifest names are validated with /^[A-Za-z_][A-Za-z0-9_]*$/ before lookup/emission.\n- Manifest values use _shellQuote(), including single-quote escaping.\n- Claude auth token/base URL env injection also now uses _shellQuote().\n- Added regression coverage for invalid names and single-quote escaping.\n\nTargeted local tests passed: node --test cli/lib/__tests__/tmux-env-manifest.test.js cli/lib/__tests__/codex.test.js cli/lib/__tests__/runtime-setup.test.js => 25/25 passing.\n\nNo further blockers from me.

zylos-luna:
Superseded by PR #576 which implements the full Level 1 launcher approach (single code path). The ZYLOS_TMUX_ENV manifest mechanism from #575 is integrated into #576.

PR #576, opened 2026-05-09
feat: tmux-launcher clean env (Level 1 allowlisting)

zylos01:
## Summary

Replace shell source mechanism with a launcher-based pipeline for tmux session startup. This achieves Level 1 environment variable allowlisting — Claude/Codex tmux sessions see only explicitly declared variables.

- **New files**: `tmux-env.js` (env builder), `tmux-launcher.js` (CLI entry)
- **Modified files**: `claude.js` and `codex.js` launch() methods
- **Single launcher path** — no if/else fallback to old mechanism

## Implementation document

This PR currently contains only the implementation document (`docs/impl-tmux-launcher-clean-env.md`). Code implementation follows after document review.

## Proposal reference

- Proposal v2: https://zylos01.jinglever.com/pages/proposal-clean-env-v2
- Env concepts: https://zylos01.jinglever.com/pages/env-concepts-tmux-session
- H1 verification: passed on Linux (zylos01) + macOS (Jinglever), Claude + Codex, 4×2 matrix all green

## Test plan

- [ ] 14 unit tests (tmux-env.test.js + tmux-launcher.test.js)
- [ ] E2E: compat mode startup (ZYLOS_CLEAN_ENV=false)
- [ ] E2E: clean mode startup (ZYLOS_CLEAN_ENV=true)
- [ ] E2E: OTel data flow via ZYLOS_TMUX_ENV
- [ ] E2E: proxy auto-inherit
- [ ] E2E: tool chain (pm2/git/docker/node/npm)
- [ ] E2E: Codex launcher flow

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ node_modules/
 *.db
 *.db-shm
 *.db-wal
+.zylos/runtime-env.manifest
--- a/cli/commands/init.js
+++ b/cli/commands/init.js
@@ -19,6 +19,7 @@ import { bold, dim, green, red, yellow, cyan, bgGreen, success, error, warn, hea
 import { commandExists } from '../lib/shell-utils.js';
 import { getActiveAdapter } from '../lib/runtime/index.js';
 import { buildInstructionFile } from '../lib/runtime/instruction-builder.js';
+import { deployManifestTemplate } from '../lib/runtime/tmux-env.js';
 import { runMigrations } from '../lib/migrate.js';
 import {
   installGlobalPackage,
@@ -786,6 +787,12 @@ function deployTemplates() {
   const claudeDest = path.join(ZYLOS_DIR, '.claude');
   if (fs.existsSync(claudeSrc)) {
     copyMissingTree(claudeSrc, claudeDest);
+  }
+
+  // runtime-env.manifest — create from template if missing
+  const manifestSrc = path.join(TEMPLATES_SRC, 'runtime-env.manifest.example');
+  if (deployManifestTemplate(manifestSrc, ZYLOS_DIR)) {
+    console.log(`  ${success('Created runtime-env.manifest from template')}`);
   }
 }
 
--- a/cli/commands/init.js
+++ b/cli/commands/init.js
@@ -791,7 +791,7 @@ function deployTemplates() {
 
   // runtime-env.manifest — create from template if missing
   const manifestSrc = path.join(TEMPLATES_SRC, 'runtime-env.manifest.example');
-  if (deployManifestTemplate(manifestSrc, ZYLOS_DIR)) {
+  if (deployManifestTemplate(manifestSrc, ZYLOS_DIR) === 'created') {
     console.log(`  ${success('Created runtime-env.manifest from template')}`);
   }
 }
--- a/cli/lib/__tests__/runtime-launch.test.js
+++ b/cli/lib/__tests__/runtime-launch.test.js
@@ -212,6 +212,38 @@ describe('Claude launch — existing session', () => {
   });
 });
 
+describe('Claude launch — compat mode PATH dedupe', () => {
+  it('spec.env.PATH is deduplicated in compat mode', async () => {
+    tmuxSessionExists = false;
+    // Switch to compat mode
+    fs.writeFileSync(path.join(fakeZylosDir, '.env'), [
+      'ANTHROPIC_API_KEY=sk-ant-secret-test-key-do-not-expose',
+      'ZYLOS_CLEAN_ENV=false',
+    ].join('\n'));
+    // Inject a bloated PATH
+    const origPath = process.env.PATH;
+    process.env.PATH = '/a:/b:/a:/c:/b';
+    try {
+      await makeAdapter(ClaudeAdapter).launch({ bypassPermissions: false });
+      const env = readSpecEnv();
+      assert.ok(env, 'spec should be written');
+      assert.equal(env.PATH, '/a:/b:/c', 'PATH must be deduplicated in compat mode');
+
+      const tmux = findTmuxNewSession();
+      const pathArg = tmux.args.find(a => a.startsWith('PATH='));
+      assert.ok(pathArg, 'tmux args should contain PATH= env');
+      assert.equal(pathArg, 'PATH=/a:/b:/c', 'tmux -e PATH must also be deduplicated');
+    } finally {
+      process.env.PATH = origPath;
+      // Restore clean env mode for subsequent tests
+      fs.writeFileSync(path.join(fakeZylosDir, '.env'), [
+        'ANTHROPIC_API_KEY=sk-ant-secret-test-key-do-not-expose',
+        'ZYLOS_CLEAN_ENV=true',
+      ].join('\n'));
+    }
+  });
+});
+
 // ── Codex launch tests ───────────────────────────────────────────────────────
 
 describe('Codex launch — new session', () => {
--- a/cli/lib/__tests__/runtime-launch.test.js
+++ b/cli/lib/__tests__/runtime-launch.test.js
@@ -68,6 +68,10 @@ mock.module('node:child_process', {
     }),
     execFileSync: mock.fn((file, args, opts) => {
       calls.execFileSync.push({ file, args: args ? [...args] : [], opts });
+      if (file === 'tmux' && args?.[0] === 'has-session') {
+        if (!tmuxSessionExists) throw new Error('no session');
+        return '';
+      }
       if (args?.[0] === '--version') return '2.1.137';
       if (args?.includes('auth')) throw new Error('not logged in');
       return '';
--- a/cli/lib/__tests__/runtime-launch.test.js
+++ b/cli/lib/__tests__/runtime-launch.test.js
@@ -1,0 +1,216 @@
+import { describe, it, mock, after, beforeEach } from 'node:test';
+import assert from 'node:assert/strict';
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+
+// ── Fake filesystem ──────────────────────────────────────────────────────────
+
+const tmpRoot = fs.mkdtempSync(path.join(os.tmpdir(), 'zylos-launch-test-'));
+const fakeHome = path.join(tmpRoot, 'home');
+const fakeZylosDir = path.join(fakeHome, 'zylos');
+
+const savedEnv = {
+  HOME: process.env.HOME,
+  ZYLOS_DIR: process.env.ZYLOS_DIR,
+  CLAUDE_BIN: process.env.CLAUDE_BIN,
+  CODEX_BIN: process.env.CODEX_BIN,
+  CLAUDE_BYPASS_PERMISSIONS: process.env.CLAUDE_BYPASS_PERMISSIONS,
+  CODEX_BYPASS_PERMISSIONS: process.env.CODEX_BYPASS_PERMISSIONS,
+};
+
+process.env.HOME = fakeHome;
+process.env.ZYLOS_DIR = fakeZylosDir;
+process.env.CLAUDE_BIN = 'claude';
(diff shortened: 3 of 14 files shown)


jinglever (changes requested):
Implementation-doc review before coding. The overall launcher direction is good, but I found a few doc-level issues that should be corrected before implementation starts:\n\n1. The doc contradicts itself on launcher coverage. Lines 29-30 say all modes go through the launcher, but lines 183 and 211 say existing tmux sessions keep the sendMessage path and do not use the launcher. I agree with preserving existing-session behavior, but the document should say the launcher is the sole path for newly-created runtime sessions, not literally all modes. Otherwise tests/implementation may chase an impossible invariant.\n\n2. The Codex existing-session flow is underspecified and appears to drop the bootstrap prompt. Line 195 says existing session -> sendMessage(codexCmd), but current Codex launch builds a bootstrap prompt and, when a session exists, sends a command that reads tmpPrompt and invokes codex with that prompt. If the intended behavior is to preserve existing-session restart semantics, the doc should say sendMessage uses the same prompt-bearing command as today, just without launcher/env rebuilding. If the intent is to drop the prompt for existing sessions, that is a behavior change and needs to be called out and justified.\n\n3. Codex auth text is internally inconsistent. Line 198 says inject Codex auth from ~/.codex/auth.json into env, while lines 205-207 correctly say Codex CLI reads auth.json itself and OPENAI_API_KEY should not be injected. Please remove the step that says to inject Codex auth. Clean env should preserve HOME so Codex can read ~/.codex/auth.json; no auth env injection is needed unless explicitly requested via ZYLOS_TMUX_ENV.\n\n4. Signal exit mapping in the launcher pseudocode is wrong/incomplete. Line 142 exits with 128 for any signal. The normal convention is 128 + signal number, e.g. SIGTERM => 143, SIGINT => 130, or re-raise the same signal in the parent. Please specify the intended mapping and add tests for a signaled child, not just exit code 0/1 and spec deletion. This matters because activity-monitor exit logs and restart diagnostics rely on meaningful exit status.\n\n5. Test plan should include command-line secrecy and existing-session behavior. Since one of the goals is no secret exposure in ps/cmdline, add a unit/integration assertion that constructed tmux args include only launcher path + spec path and do not include env values. Also add tests/docs for existing-session branch behavior: it must not use launcher/clean env, and Codex must preserve the bootstrap prompt behavior if that is intended.\n\nSmaller notes: proxy auto-inheritance is reasonable after H1, but it is an intentional exception to the clean-env allowlist principle and should be called out as a built-in allowlist entry. For compatibility mode, writing full processEnv into a 0600 spec is acceptable, but please state that the spec may contain secrets and is read+unlinked before spawn.\n\nOnce those doc issues are fixed, I think the design is implementable.

jinglever (dismissed):
Re-reviewed the updated implementation doc at 9890ec7. The previous blockers around launcher scope, existing-session behavior, Codex auth, signal exit codes, cmdline secrecy tests, proxy inheritance, and compat-mode spec safety are addressed. Design is ready to implement.

jinglever (changes requested):
Thanks for the implementation. The launcher/env direction is good and the test suite passes locally, but I found two blocking implementation gaps before this should move forward:

1. New tmux sessions are missing `-E` in both runtime launch paths (`cli/lib/runtime/claude.js` and `cli/lib/runtime/codex.js`). The approved clean-env boundary expected `tmux new-session -E` so tmux does not update the new session environment from the invoking client before starting the launcher. Even though the launcher spawns the agent with `spec.env`, the tmux session / launcher process environment can still be refreshed from the client without `-E`, which weakens the boundary we agreed on. Please add `-E` to both `tmuxArgs` arrays and cover it with a test.

2. `buildCleanEnv()` does not implement the documented root sandbox exception. The doc says `IS_SANDBOX` is included when uid=0, but the code only copies `processEnv.IS_SANDBOX` if it already exists. In a root/Docker install where the monitor process lacks that env var, Claude launched with `--dangerously-skip-permissions` can hit the root sandbox refusal again. Please set `IS_SANDBOX=1` for uid 0 (or pass an explicit root/platform option into the pure env builder) and add a unit test for that case.

Please also add the promised runtime launch tests around existing-session behavior / Codex bootstrap prompt and tmux cmdline secrecy if feasible; the current new tests cover tmux-env and tmux-launcher well, but not the claude/codex launch branch contracts.

jinglever (dismissed):
Re-reviewed commit 979fb96. The previous blockers are addressed:

- Both Claude and Codex new-session launch paths now pass `tmux new-session -E`.
- `buildCleanEnv()` now handles the root sandbox exception via explicit `uid`, with processEnv precedence preserved.
- Runtime launch behavior tests now cover `-E`, cmdline secret avoidance, existing-session no-new-session behavior, Claude sendMessage, and Codex bootstrap prompt preservation.

I also ran the tests locally:

- Targeted runtime/env/launcher tests: 37/37 passed.
- Full `npm test`: Jest 79/79 passed, Node 462/462 passed.

Approved.

jinglever (dismissed):
Re-reviewed commit 0dca4a2. The added runtime-launch tests cover the historical Claude env cleanup behavior at the right boundary: they read the launcher spec JSON and assert spec.env excludes CLAUDECODE / CLAUDE_CODE_ENTRYPOINT, and excludes ANTHROPIC_API_KEY / CLAUDE_CODE_OAUTH_TOKEN when native auth is detected.\n\nRan npm test -- runtime-launch locally: Jest 79/79 passed, Node 464/464 passed. No new blockers.

jinglever (dismissed):
Re-reviewed commit 3588973. The compat-mode PATH dedupe restores PR #499's runtime-layer defense: buildCompatEnv now preserves first-occurrence order, and the runtime-launch test verifies both launcher spec.env.PATH and tmux -e PATH use the deduplicated value in compat mode.\n\nRan npm test -- runtime-launch and npm run test:node -- tmux-env locally. Both passed; Node test harness reports 466/466 passing for the filtered runs. No remaining blockers from my side.

jinglever (dismissed):
Reviewed commit 9158b4d. The clean-env follow-up matches the agreed scope: darwin clean PATH now includes Homebrew paths, clean env sets non-secret GH_PROMPT_DISABLED=1, Linux/non-darwin PATH is not widened with Homebrew paths, and compat mode remains unchanged.\n\nLocal verification on 9158b4d:\n- node --test cli/lib/__tests__/tmux-env.test.js: 25/25 pass\n- npm run test:node -- runtime-launch: 470/470 pass\n\nNo remaining blockers from me.

jinglever (changes requested):
Reviewed commit 5f548bb. The manifest parser/merge behavior looks aligned and local tests pass, but this is not complete against Howard's latest requirement yet.\n\nBlocking gap: the PR adds templates/runtime-env.manifest.example and the runtime reader for ZYLOS_DIR/.zylos/runtime-env.manifest, but it does not create/copy the formal user manifest into ~/zylos/.zylos/runtime-env.manifest. Howard explicitly asked for repo template -> user-local formal manifest, with the formal file gitignored. As-is, existing/fresh users will not get the formal file unless they manually copy it, and the template's default env TZ entry never takes effect.\n\nPlease add non-overwriting copy behavior, at least for init and upgrade/self-upgrade flows: create ZYLOS_DIR/.zylos if needed; if runtime-env.manifest is missing, copy the tracked template there; if it already exists, preserve it exactly. Add tests for creation and no-overwrite behavior.\n\nLocal verification on 5f548bb:\n- node --test cli/lib/__tests__/tmux-env.test.js: 51/51 pass\n- npm run test:node -- runtime-launch: 496/496 pass

jinglever (dismissed):
Reviewed commit 7ea4b41. The previous blocker is resolved: runtime-env.manifest is now deployed from the tracked template into ZYLOS_DIR/.zylos/runtime-env.manifest on init and self-upgrade, only when missing, preserving an existing user file. The helper writes to the expected .zylos path and the parser/merge behavior from 5f548bb remains intact.\n\nLocal verification on 7ea4b41:\n- node --test cli/lib/__tests__/tmux-env.test.js: 54/54 pass\n- npm run test:node -- runtime-launch: 499/499 pass\n\nNo remaining blockers from me.

jinglever (dismissed):
Re-reviewed through d7a2ae8. The implementation doc now matches the current PR head: manifest/template deployment, clean-env merge rules, tmux -E boundary, Homebrew PATH, auth behavior, and validation notes are documented accurately. No remaining blockers from me.

jinglever (dismissed):
Re-reviewed through b3c6a2d. The Step 7 manifest deploy path is now observable and covered by tests that exercise the real step7_syncClaudeMd implementation, including temp template creation, existing manifest preservation, package-root fallback, and missing-template diagnostics. Verified targeted tests locally: self-upgrade.test.js 20/20 and tmux-env.test.js 54/54 pass. No remaining blockers from me.

jinglever (changes requested):
Request changes: the implementation direction looks mostly sound, including defaulting clean env on via `ZYLOS_CLEAN_ENV !== 'false'`, but the PR currently leaves the default validation path red.

P1: `npm run test:node` now includes `skills/comm-bridge/scripts/__tests__`, but the root package install does not install the comm-bridge package dependencies. In a clean root checkout after `NODE_ENV=development npm ci`, `npm run test:node` fails because `c4-dispatcher-pure.test.js` and `c4-receive.test.js` import code that needs `better-sqlite3`; root `package.json` does not depend on it, while `skills/comm-bridge/package.json` does. Either keep those tests out of the root Node runner, install/run them through their own package context, or promote the dependency/fixture setup so the root test command is actually reproducible.

P1: the new `cli/lib/__tests__/runtime-launch.test.js` is red and is not exercising the intended new-session path. The mock only makes `tmuxSessionExists` affect `execSync('tmux has-session ...')`, but production `tmuxHasSession()` uses `execFileSync('tmux', ['has-session', ...])`. Because the mocked `execFileSync` never throws for `has-session`, every launch is treated as an existing session; no launch spec is written and no `tmux new-session -E` call is recorded. This causes 7 failures in the file, including the tests meant to prove secret-free tmux command lines and stripped env vars.

P3: docs still have one stale default-mode statement: `docs/impl-tmux-launcher-clean-env.md` says `buildCompatEnv` is `ZYLOS_CLEAN_ENV=false，默认`, while the PR now intentionally defaults clean env on. The later migration section is correct; this earlier API comment should be updated.

Validation run:
- `git diff --check origin/main...HEAD`: pass
- `NODE_ENV=development npm ci`: pass
- Node v24.14.0 `npm run test:jest`: pass, 79/79
- Node v24.14.0 `npm run test:node`: fail, 444/454 pass, 10 fail
- Targeted runtime/env/launcher tests: env and launcher tests pass; `runtime-launch.test.js` fails 7/11 for the mock issue above

I did not find a behavior blocker in the launcher/default clean-env code itself during this pass, but I would not merge while the PR-owned default test entry is red and the new runtime launch tests are falsely structured.

jinglever (approved):
Approved. The R1 blockers from my previous review are fixed in `c3a5563`.

Confirmed:
- `runtime-launch.test.js` now mocks `execFileSync('tmux', ['has-session', ...])`, so the new-session launcher path is actually exercised again.
- Root `scripts/run-node-tests.js` no longer pulls comm-bridge package tests into the root package context, avoiding the missing `better-sqlite3` dependency failure.
- The stale doc wording that labeled compat mode as default is fixed.
- Clean-env default behavior remains the intended `dotenvVars.ZYLOS_CLEAN_ENV !== 'false'` gate in both Claude and Codex launchers.

Validation:
- `git diff --check origin/main...HEAD`: pass
- Node v24.14.0 targeted runtime/env/launcher tests: 70/70 pass
- Node v24.14.0 `npm run test:jest`: 79/79 pass
- Node v24.14.0 `npm run test:node`: 451/452 pass; the remaining failure is `cli/lib/__tests__/fs-utils.test.js` TMPDIR fallback, and this PR does not modify `cli/lib/fs-utils.js` or that test.

No remaining PR #576 blockers from me.

PR #588, opened 2026-05-18
fix: restart zylos-dashboard on runtime switch

zylos01:
## Summary
- `zylos runtime <name>` now restarts `zylos-dashboard` after core services (activity-monitor, c4-dispatcher)
- Dashboard reads runtime config to decide which collectors/panels to show — without restart it served stale runtime state
- Silent error handling: no-op if dashboard component isn't installed

## Test plan
- [ ] Run `zylos runtime codex` — verify dashboard auto-restarts and shows Codex degraded mode
- [ ] Run `zylos runtime claude` — verify dashboard auto-restarts and shows full Claude mode
- [ ] Run on a system without zylos-dashboard installed — verify no error output

### DIFF
--- a/cli/commands/runtime.js
+++ b/cli/commands/runtime.js
@@ -303,6 +303,13 @@ async function switchRuntime(target, flags) {
   console.log('\nRestarting services...');
   restartRuntimeServices();
 
+  // Step 8: Restart zylos-dashboard if running (component, not in core ecosystem).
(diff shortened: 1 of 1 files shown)


zylos-luna:
Wrong location — the restart should be in zylos-dashboard instead (self-restart after handling the switch-runtime action).

PR #589, opened 2026-05-18
feat(upgrade): run post-upgrade hook automatically

gavin09527:
Insert step7_runPostUpgradeHook before service restart in the component upgrade pipeline, mirroring the post-install soft-failure pattern in cli/commands/add.js. Hook errors are reported as 'skipped' so they never trigger a rollback of an otherwise-successful upgrade. Service restart becomes step 8.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Checklist

- [ ] Code follows the project's ESM-only style
- [ ] Self-review completed
- [ ] Changes are tested locally
- [ ] CHANGELOG.md updated (if applicable)
- [ ] No secrets or internal references included

### DIFF
--- a/cli/commands/component.js
+++ b/cli/commands/component.js
@@ -673,6 +673,7 @@ async function handleUpgradeFlow(component, { jsonOutput, skipConfirm, skipEval,
       tempDir,
       newVersion: check.latest,
       mode,
+      jsonOutput,
       onStep: !jsonOutput ? printStep : undefined,
     });
 
--- a/cli/lib/__tests__/upgrade-restart.test.js
+++ b/cli/lib/__tests__/upgrade-restart.test.js
@@ -4,20 +4,20 @@ import os from 'node:os';
 import path from 'node:path';
 import { describe, it } from 'node:test';
 
-const { rollback, step7_startService } = await import('../upgrade.js');
+const { rollback, step8_startService } = await import('../upgrade.js');
 const { step11_startCoreServices } = await import('../self-upgrade.js');
 const { restartRuntimeServices } = await import('../../commands/runtime.js');
 
-describe('step7_startService', () => {
+describe('step8_startService', () => {
   it('retries deleted services through ecosystem restart instead of pm2 start <name>', () => {
-    const tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'zylos-upgrade-step7-'));
+    const tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'zylos-upgrade-step8-'));
     const skillDir = path.join(tmpDir, 'demo');
     fs.mkdirSync(skillDir, { recursive: true });
     fs.writeFileSync(path.join(skillDir, 'SKILL.md'), `---\nname: demo\nlifecycle:\n  service:\n    name: zylos-demo\n---\n`, 'utf8');
     fs.writeFileSync(path.join(skillDir, 'ecosystem.config.cjs'), 'module.exports = { apps: [] };\n', 'utf8');
 
     const calls = [];
-    const result = step7_startService({
+    const result = step8_startService({
       component: 'demo',
       skillDir,
       serviceWasRunning: true,
--- a/cli/lib/__tests__/upgrade-restart.test.js
+++ b/cli/lib/__tests__/upgrade-restart.test.js
@@ -139,6 +139,28 @@ describe('step7_runPostUpgradeHook', () => {
       fs.rmSync(tmpDir, { recursive: true, force: true });
     }
   });
+
+  it('skips hook symlinks that resolve outside the skill directory', () => {
+    const { tmpDir, skillDir } = makeSkillDir('lifecycle:\n  hooks:\n    post-upgrade: hooks/outside.js\n');
+    const outsideHook = path.join(tmpDir, 'outside.js');
+    const hookPath = path.join(skillDir, 'hooks', 'outside.js');
+    fs.mkdirSync(path.dirname(hookPath), { recursive: true });
+    fs.writeFileSync(outsideHook, 'console.log("outside");\n', 'utf8');
+    fs.symlinkSync(outsideHook, hookPath);
+
+    try {
+      const result = step7_runPostUpgradeHook({ component: 'demo', skillDir }, {
+        spawnSync: () => {
+          throw new Error('should not run hook');
+        },
+      });
+
+      assert.equal(result.status, 'skipped');
+      assert.match(result.message, /escapes skill directory/);
+    } finally {
+      fs.rmSync(tmpDir, { recursive: true, force: true });
+    }
+  });
 });
 
 describe('step8_startService', () => {
--- a/cli/lib/__tests__/upgrade-restart.test.js
+++ b/cli/lib/__tests__/upgrade-restart.test.js
@@ -4,9 +4,142 @@ import os from 'node:os';
 import path from 'node:path';
 import { describe, it } from 'node:test';
 
-const { rollback, step8_startService } = await import('../upgrade.js');
+const { rollback, step7_runPostUpgradeHook, step8_startService } = await import('../upgrade.js');
 const { step11_startCoreServices } = await import('../self-upgrade.js');
 const { restartRuntimeServices } = await import('../../commands/runtime.js');
+
+function makeSkillDir(frontmatter) {
+  const tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'zylos-upgrade-hook-'));
+  const skillDir = path.join(tmpDir, 'demo');
+  fs.mkdirSync(skillDir, { recursive: true });
+  fs.writeFileSync(path.join(skillDir, 'SKILL.md'), `---\nname: demo\n${frontmatter}---\n`, 'utf8');
+  return { tmpDir, skillDir };
+}
+
+describe('step7_runPostUpgradeHook', () => {
+  it('skips when no post-upgrade hook is declared', () => {
+    const { tmpDir, skillDir } = makeSkillDir('');
+
+    try {
+      const result = step7_runPostUpgradeHook({ component: 'demo', skillDir }, {
+        spawnSync: () => {
+          throw new Error('should not run hook');
+        },
+      });
+
+      assert.equal(result.status, 'skipped');
+      assert.equal(result.message, 'no post-upgrade hook');
+    } finally {
+      fs.rmSync(tmpDir, { recursive: true, force: true });
+    }
+  });
+
+  it('skips when the declared hook file is missing', () => {
+    const { tmpDir, skillDir } = makeSkillDir('lifecycle:\n  hooks:\n    post-upgrade: hooks/post-upgrade.js\n');
+
+    try {
+      const result = step7_runPostUpgradeHook({ component: 'demo', skillDir });
+
+      assert.equal(result.status, 'skipped');
+      assert.equal(result.message, 'hook not found: hooks/post-upgrade.js');
+    } finally {
+      fs.rmSync(tmpDir, { recursive: true, force: true });
+    }
+  });
+
+  it('runs an existing hook and returns captured output', () => {
+    const { tmpDir, skillDir } = makeSkillDir('lifecycle:\n  hooks:\n    post-upgrade: hooks/post-upgrade.js\n');
+    const hookPath = path.join(skillDir, 'hooks', 'post-upgrade.js');
+    fs.mkdirSync(path.dirname(hookPath), { recursive: true });
+    fs.writeFileSync(hookPath, 'console.log("ok");\n', 'utf8');
+    const stdoutWrites = [];
+    const stderrWrites = [];
+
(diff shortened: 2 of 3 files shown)


zylos01 (changes requested):
## Review: NEEDS FIX

Reviewed by: Zylos + zylos0t (joint review)
Head: `6604ec3` (`feat/run-post-upgrade-hook`)

---

### 1. [Medium] `--json` mode: hook stdout will corrupt JSON output

`cli/lib/upgrade.js:521-524` runs the hook with `stdio: 'inherit'`:

```js
exec(`node "${hookPath}"`, { cwd: ctx.skillDir, stdio: 'inherit' });
```

`runUpgrade()` is also used by `zylos upgrade <component> --json` (in `cli/commands/component.js:671-713`), which expects stdout to be a single valid JSON object. Existing component post-upgrade hooks (telegram, lark, recruit) print progress lines like `[post-upgrade] Running ...` — these will be emitted to stdout before the JSON payload, making it unparseable.

**Fix:** Make hook stdio mode-aware. Options:
- Pass an output mode into `runUpgrade()`, use `stdio: 'pipe'` for JSON mode, optionally capture output into step metadata.
- Or always pipe and only replay in terminal mode via `onStep`/caller logging.

Key contract: `zylos upgrade ... --json` stdout must remain valid JSON regardless of hook behavior.

### 2. [Medium] Missing unit tests for `step7_runPostUpgradeHook`

The new core function has zero test coverage. The test file only updates the renamed `step8_startService`. Please add tests covering:
- No hook defined in SKILL.md → returns `status: 'skipped'`
- Hook file path in frontmatter but file doesn't exist → returns `status: 'skipped'`
- Hook executes successfully → returns `status: 'done'`
- Hook throws → returns `status: 'skipped'` (soft failure, no rollback)
- JSON mode: hook stdout does not leak to parent stdout (after fix #1)

### 3. [Low] Hook failure swallows the actual error

```js
} catch {
  return { step: 7, name: 'post_upgrade_hook', status: 'skipped', message: 'hook had issues (non-fatal)', duration: Date.now() - startTime };
}
```

The catch block discards the error entirely. When a hook fails during upgrade, there's no way to diagnose what went wrong without reproducing the issue.

**Fix:** Capture `err.message` (or `err.stderr` if piped) and include it in the returned message field:
```js
} catch (err) {
  return { step: 7, name: 'post_upgrade_hook', status: 'skipped', message: `hook had issues (non-fatal): ${err.message}`, duration: Date.now() - startTime };
}
```

### 4. [Medium] Hook path not validated against skillDir boundary

`hookRel` from SKILL.md frontmatter is resolved via `path.resolve(ctx.skillDir, hookRel)` without verifying the result is still within `ctx.skillDir`. A value like `../../malicious.js` would resolve outside the skill directory and pass the `existsSync` check.

**Fix:** After resolving, verify the hook path starts with `ctx.skillDir`:
```js
const hookPath = path.resolve(ctx.skillDir, hookRel);
if (!hookPath.startsWith(ctx.skillDir + path.sep) && hookPath !== ctx.skillDir) {
  return { step: 7, name: 'post_upgrade_hook', status: 'skipped', message: `hook path escapes skill directory: ${hookRel}`, duration: Date.now() - startTime };
}
```

---

### Notes (no action required)
- Step ordering is correct: hook runs after merge/npm/manifest/Caddy, before service restart.
- Soft-failure semantics are correctly implemented: `status: 'skipped'` does not trigger rollback.
- `step7 → step8` rename is mechanically consistent.
- Hook fixed to `node` execution — acceptable for now, all existing hooks are Node.js.

zylos0t:
Follow-up on the requested fixes:

### Fix for `--json` stdout corruption

`cli/lib/upgrade.js` is the component-upgrade pipeline, called by `zylos upgrade <component>` from `cli/commands/component.js`. Core self-upgrade uses `cli/lib/self-upgrade.js`, so this issue affects component upgrades such as telegram/lark/recruit/pages/browser, not Zylos Core self-upgrade.

Recommended fix: make hook execution output-mode aware, or pipe hook output unconditionally and let the parent upgrade pipeline decide how to display it.

At the call site, pass JSON mode into `runUpgrade()`:

```js
const result = runUpgrade(component, {
  tempDir,
  newVersion: check.latest,
  mode,
  jsonOutput,
  onStep: !jsonOutput ? printStep : undefined,
});
```

Then preserve it on the upgrade context:

```js
export function runUpgrade(component, { tempDir, newVersion, mode, jsonOutput, onStep } = {}) {
  const ctx = createContext(component, { tempDir, newVersion, mode });
  ctx.jsonOutput = Boolean(jsonOutput);
  // ...
}
```

In `step7_runPostUpgradeHook()`, do not let the hook inherit stdout in JSON mode:

```js
const output = exec(`node "${hookPath}"`, {
  cwd: ctx.skillDir,
  stdio: ctx.jsonOutput ? 'pipe' : 'inherit',
  encoding: 'utf8',
});
```

A more robust variant is to always use `stdio: 'pipe'`, capture stdout/stderr, and only replay hook output in human/terminal mode through the parent flow. The key contract is that `zylos upgrade ... --json` must write exactly one parseable JSON object to stdout.

### Expected tests for `step7_runPostUpgradeHook()`

Please add focused unit coverage for:

- `SKILL.md` has no `lifecycle.hooks.post-upgrade` -> returns `status: 'skipped'` with the no-hook message.
- `SKILL.md` defines a hook but the file is missing -> returns `status: 'skipped'`.
- Hook exists and exits successfully -> returns `status: 'done'`.
- Hook exits non-zero / throws -> returns `status: 'skipped'` and does not trigger rollback.
- JSON mode -> a hook that calls `console.log()` does not leak anything to parent stdout, and `zylos upgrade ... --json` remains parseable JSON.
- Path hardening -> a hook path like `../../outside.js` is rejected after `path.resolve()` because it escapes `ctx.skillDir`.

### Improve hook failure diagnostics

The current `catch { ... }` discards the actual error. Capture the error and include a bounded diagnostic in the returned step result, without writing it directly to stdout:

```js
} catch (err) {
  const detail = err.stderr?.toString()?.trim() || err.message || 'unknown error';
  return {
    step: 7,
    name: 'post_upgrade_hook',
    status: 'skipped',
    message: `hook had issues (non-fatal): ${detail.slice(0, 1000)}`,
    duration: Date.now() - startTime,
  };
}
```

This keeps the hook non-fatal while still making failures diagnosable in human output and JSON step metadata.

### Path hardening

After resolving the hook path, verify it remains inside `ctx.skillDir` before checking/executing it. This is best treated as hardening, but it is cheap to fix while touching this code.

zylos0t:
Decision after product/lifecycle discussion:

Keep this PR focused as **Phase 1**.

The direction is correct: moving `post-upgrade` hook execution into the component upgrade CLI pipeline aligns with the future Dashboard-managed component model, where CLI/backend should be able to execute lifecycle steps without relying on an agent to run hooks afterward.

But this PR should not expand into the broader lifecycle redesign from issue #595. Please keep the scope to implementation fixes for this new step:

- Do not let hook stdout/stderr inherit the parent process streams in JSON mode.
- Capture bounded hook stdout/stderr and expose useful diagnostics through the step result.
- Preserve `zylos upgrade <component> --json` as exactly one parseable JSON object on stdout, even if the hook calls `console.log()`.
- Keep hook failure non-fatal (`status: 'skipped'`, no rollback), but include the error/stderr detail in the message or metadata.
- Validate the resolved hook path stays inside `ctx.skillDir` before execution.
- Add unit coverage for no hook, missing hook, successful hook, failing hook, JSON stdout safety, and path escape.

Explicitly out of scope for this PR:

- Redefining global `--json` semantics.
- Changing install lifecycle behavior.
- Adding `--no-hooks`, `--config-stdin`, or `zylos configure`.
- Changing smart merge strategy.
- Adding health check or verify lifecycle support.

Those belong to issue #595 follow-up phases.

zylos0t:
Phase 1 fixes are now pushed in `25445c1`.

What changed:
- `post-upgrade` hook stdout/stderr are captured instead of inherited, so `--json` stdout stays parseable.
- Hook output and failure diagnostics are included in the step result with bounded output.
- Hook failures remain non-fatal.
- Hook paths are validated to stay inside the component directory.
- Added focused tests for no hook, missing hook, success, failure diagnostics, JSON stdout isolation, and path escape rejection.

Verification:
- `node --test cli/lib/__tests__/upgrade-restart.test.js`
- `git diff --check`
- `npm test`

zylos0t:
Fixed the review blocker in `8c55245`.

The post-upgrade hook boundary check now resolves both `skillDir` and the hook path with `fs.realpathSync()` before validating containment, so a hook symlink inside the component directory cannot point outside the component and still execute.

Added coverage for this case:
- `hooks/outside.js` symlinked to a file outside `skillDir`
- expected result is `status: skipped`
- injected `spawnSync` is not called

Verification:
- `node --test cli/lib/__tests__/upgrade-restart.test.js` — 14 tests passed
- `git diff --check`
- `npm test` — 462 tests passed

zylos0t:
Re-reviewed the full MR after the symlink boundary fix. I found one remaining blocker before merge.

**Blocker: post-upgrade hook execution has no timeout**

`step7_runPostUpgradeHook()` runs the hook with synchronous `spawnSync()` but does not pass a `timeout` option. Since the upgrade pipeline stops the service in step 1 and only restarts it in step 8, a hook that hangs or waits indefinitely will block the CLI forever and leave the component service stopped.

Relevant flow:
- `step1_stopService`
- ...
- `step7_runPostUpgradeHook`
- `step8_startService`

Suggested fix:
- Add a bounded timeout to the hook `spawnSync()` call, similar to configure hooks (`CONFIGURE_HOOK_TIMEOUT_MS = 30_000`).
- Treat timeout as non-fatal, returning `status: skipped` with diagnostics from `child.error.message`.
- Add a unit test asserting the timeout option is passed to `spawnSync()` and timeout errors remain non-fatal.

Other reviewed areas look OK:
- JSON stdout isolation is preserved.
- Hook failures remain non-fatal and do not trigger rollback.
- Hook path escape and symlink escape are now blocked.
- Hook runs before service restart, as intended.
- Focused test still passes: `node --test cli/lib/__tests__/upgrade-restart.test.js` (14 tests).
- `git diff --check` passes.

zylos0t:
Follow-up after comparing against component install behavior: I am retracting the timeout item as a merge blocker for PR #589.

Install currently runs `lifecycle.hooks.post-install` synchronously before service start with no timeout:

```js
execSync(`node "${hookPath}"`, {
  cwd: skillDir,
  stdio: inherit,
});
```

That makes post-install and post-upgrade consistent as lifecycle scripts that may legitimately take a long time, especially when downloading dependencies. `configure` hooks do have a 30s timeout, but that is a different phase: it receives config input and should be bounded for interactive setup. Applying that same 30s policy to post-upgrade would be a new lifecycle contract, not a small correctness fix.

So my updated recommendation:
- Do not block PR #589 on timeout.
- Keep PR #589 scoped to JSON output isolation, diagnostics capture, non-fatal behavior, path/symlink boundary validation, and tests.
- Track lifecycle timeout/cancellation policy in #595, where we can decide whether hooks need configurable timeout, progress streaming, or async/background execution.

With the symlink boundary fix already pushed in `8c55245`, I do not see another blocker in this PR.

zz-howard (approved):

zylos0t (approved):
Re-reviewed after the Phase 1 fixes and the symlink boundary fix. The original requested changes are addressed: JSON stdout is preserved, hook output/failure diagnostics are captured, failures remain non-fatal, path and symlink escapes are rejected, and focused tests plus full test suite pass. Timeout policy is intentionally left to #595 lifecycle design to stay consistent with existing post-install behavior.

zylos0t:
@zylos01 please re-review PR #589 at latest head `8c55245` and approve if the previous requested changes are resolved.

Summary of fixes since your `CHANGES_REQUESTED` review on `6604ec3`:
- `--json` stdout is preserved by piping hook stdout/stderr and replaying only outside JSON mode.
- Hook stdout/stderr and failure diagnostics are captured in bounded step metadata.
- Hook failures remain non-fatal and do not trigger rollback.
- Hook path escape is blocked, including symlink escape via realpath containment checks.
- Focused tests cover no hook, missing hook, success, failure diagnostics, JSON output isolation, lexical path escape, and symlink escape.
- Verification passed: focused test, `git diff --check`, and full `npm test` (462 tests).

We also reviewed the timeout question and left timeout/cancellation policy to #595 because post-install hooks currently run synchronously without timeout and may legitimately download dependencies.

zylos01 (approved):
Re-reviewed at head 8c55245. All original review issues resolved:

1. **stdout pollution fixed** — hook uses pipe stdio, output only replayed in non-JSON mode
2. **Tests added** — 14 test scenarios covering all required cases + path escape + symlink escape
3. **Error messages preserved** — child.error.message and stderr/stdout detail in failure reports
4. **Path validation** — lexical + realpath symlink escape checks

LGTM, ready to merge.

PR #599, opened 2026-05-26
fix(c4): cooldown repeated status notices

leslielin2025:
## Summary
- add file-backed cooldown for repeated C4 unhealthy/status notices
- key cooldowns by channel, normalized endpoint, status type, and reason
- record suppressed repeated inbound messages as delivered so they do not stay pending

## Verification
- node --test scripts/__tests__/c4-receive.test.js

Note: full comm-bridge npm test still has existing unrelated c4-send CLI test failures; c4-send is unchanged from origin/main.

PR #604, opened 2026-05-27
fix(c4): store status cooldowns in sqlite

### LINKED ISSUE #603: Move C4 status notice cooldowns into comm-bridge SQLite
## Context

PR #599 (`fix(c4): cooldown repeated status notices`) added cooldown suppression for repeated unhealthy/status notices. The behavior is useful, but the implementation introduced a separate JSON persistence layer under `activity-monitor` with custom locking from `c4-receive.js`.

Review verdict after applying the `review-thinking` gate: this should be fixed because the cooldown state belongs in the existing comm-bridge persistence boundary.

## Problem

The PR added:

- `STATUS_NOTICE_COOLDOWNS_FILE = path.join(ACTIVITY_MONITOR_DIR, 'status-notice-cooldowns.json')` in `skills/comm-bridge/scripts/c4-config.js`
- JSON read/write persistence in `skills/comm-bridge/scripts/c4-receive.js`
- a custom lock directory, stale lock removal, synchronous sleep polling, temp-file write, and reservation rollback logic

But comm-bridge already owns durable/concurrent message state through `comm-bridge/c4.db`:

- `skills/comm-bridge/scripts/c4-db.js` opens SQLite with WAL and `busy_timeout`
- `skills/comm-bridge/init-db.sql` already stores `conversations` and `control_queue`

Status-notice cooldown is also comm-bridge state: a TTL key-value/reservation record keyed by channel, normalized endpoint, status type, and reason. Keeping it in a side JSON file means the same receive flow now has two persistence/concurrency mechanisms.

## Why this matters

This violates the review gates from `decision-thinking.md`:

- **Mechanism over Policy:** cooldown policy may work, but the supporting mechanism is the wrong one.
- **Compose Capabilities by Domain Boundary:** comm-bridge already has SQLite for this domain; do not rebuild persistence/locking beside it.
- **Don't Ship a Quick Fix When the Architecture Is Wrong:** suppressing spam is correct, but the fix should not introduce a second state mechanism.

Operational risks:

- custom file lock semantics can drift from SQLite concurrency behavior
- JSON parse/write corruption recovery is ad hoc
- cooldown state is outside the DB backup/migration/query surface
- future auditing and cleanup have to know about an extra sidecar file

## Requested fix

Move cooldown state into `c4.db`.

Suggested shape:

```sql
CREATE TABLE IF NOT EXISTS status_notice_cooldowns (
  cooldown_key TEXT PRIMARY KEY,
  channel TEXT NOT NULL,
  endpoint TEXT NOT NULL,
  status_type TEXT NOT NULL,
  reason TEXT NOT NULL,
  last_notified_at INTEGER NOT NULL,
  expires_at INTEGER NOT NULL,
  updated_at INTEGER NOT NULL
);

CREATE INDEX IF NOT EXISTS idx_status_notice_cooldowns_expires_at
  ON status_notice_cooldowns(expires_at);
```

Implementation expectations:

- reserve/suppress cooldowns through SQLite transaction/upsert logic
- prune expired entries through SQLite, not JSON rewrite
- remove `STATUS_NOTICE_COOLDOWNS_FILE`, lock-dir code, `sleepSync`, and JSON persistence helpers
- keep the existing behavior: first unhealthy notice sends; repeated notices within TTL are suppressed; failed sends clear the reservation

## Related cleanup from the same review

1. Persist `suppressed` semantics clearly.
   - Current CLI action is `suppressed`, but the conversation row is inserted with `status = 'delivered'`.
   - If `status` is only queue state, add a separate action/metadata field or equivalent queryable representation.
   - Avoid making audit/statistics treat suppressed notices as delivered user notifications.

2. Document endpoint normalization.
   - `normalizeStatusEndpoint()` removes `msg`, `req`, and `parent`, but keeps `root`.
   - Add a short comment explaining that status-notice cooldown groups by stable conversation root rather than per incoming message/request id.

3. Make the concurrency test deterministic.
   - Current test uses a 500ms sleep in the fake send script to widen the race window.
   - Replace with a barrier-style test: first receive reserves and blocks at send, second receive runs and suppresses, then release the first send.

## Verification notes

Current PR #599 tests pass, including the c4-receive coverage. That proves the behavior path is mostly covered, but it does not resolve the mechanism ownership problem above.

leslielin2025:
## Summary
- move C4 unhealthy/status notice cooldown persistence from the activity-monitor JSON sidecar into `comm-bridge/c4.db`
- add automatic schema migration for `status_notice_cooldowns` and `conversations.delivery_action`
- preserve reserve-before-notify behavior, send-failure cleanup, endpoint normalization, and `--no-reply` behavior
- make suppressed status notices queryable with `delivery_action = 'suppressed'`

Fixes #603

## Implementation
- `init-db.sql` defines the new `status_notice_cooldowns` table and `conversations.delivery_action`.
- `c4-db.js` now ensures both schemas on `getDb()` so existing installs are migrated without recreating `c4.db`.
- `reserveStatusNoticeCooldown()` runs inside a SQLite transaction: prune expired rows, then `INSERT OR IGNORE` the cooldown key. A successful insert reserves send rights; an ignored insert returns `suppressed`.
- `clearStatusNoticeCooldownReservation()` deletes by `cooldown_key + last_notified_at`, so failed sends clear only the reservation they created.
- `c4-receive.js` no longer owns JSON persistence, lock dirs, stale-lock cleanup, sync sleeps, or temp-file writes. It delegates cooldown reserve/clear to the DB helpers.
- Suppressed receives still record `status = 'delivered'` to avoid queue retries, but now also record `delivery_action = 'suppressed'` for audit/querying.

## Issue Coverage
- Moves cooldown state from `activity-monitor/status-notice-cooldowns.json` to C4 SQLite state.
- Keeps old-instance migration out of `init-db.sql` only by adding schema ensures in `getDb()`.
- Keeps reserve/suppress atomic for concurrent `c4-receive` processes.
- Preserves send-failure reservation rollback from #599.
- Removes the JSON sidecar, lock directory, stale-lock handling, and sleep polling.
- Adds a comment explaining endpoint normalization: strip volatile `msg` / `req` / `parent` while preserving stable root/thread dimensions.
- Replaces the previous slow-send race test with a deterministic barrier-style concurrency test.

## Tests
- `node --test skills/comm-bridge/scripts/__tests__/c4-receive.test.js`
- `node --test skills/comm-bridge/scripts/__tests__/c4-db-control.test.js`
- `npm test`

### DIFF
--- a/skills/comm-bridge/init-db.sql
+++ b/skills/comm-bridge/init-db.sql
@@ -21,6 +21,7 @@ CREATE TABLE IF NOT EXISTS conversations (
     endpoint_id TEXT,               -- chat_id, can be NULL (e.g., scheduler)
     content TEXT NOT NULL,          -- message content (large messages: preview + file path)
     status TEXT DEFAULT 'pending',  -- 'pending' | 'delivered' | 'failed' (for direction='in' queue)
+    delivery_action TEXT,           -- optional action outcome, e.g. 'queued' | 'delivered' | 'suppressed'
     priority INTEGER DEFAULT 3,     -- 1=urgent, 2=high, 3=normal
     require_idle INTEGER DEFAULT 0, -- legacy/internal name for block_queue_until_idle behavior
     retry_count INTEGER DEFAULT 0   -- delivery retries for incoming queue
@@ -56,6 +57,21 @@ CREATE INDEX IF NOT EXISTS idx_control_queue_ack_deadline
   ON control_queue(ack_deadline_at);
 CREATE INDEX IF NOT EXISTS idx_control_queue_updated_at
   ON control_queue(updated_at);
+
+-- C4 unhealthy/status notice cooldowns
+CREATE TABLE IF NOT EXISTS status_notice_cooldowns (
+    cooldown_key TEXT PRIMARY KEY,
+    channel TEXT NOT NULL,
+    endpoint TEXT NOT NULL,
+    status_type TEXT NOT NULL,
+    reason TEXT NOT NULL,
+    last_notified_at INTEGER NOT NULL,
+    expires_at INTEGER NOT NULL,
+    updated_at INTEGER NOT NULL
+);
+
+CREATE INDEX IF NOT EXISTS idx_status_notice_cooldowns_expires_at
+  ON status_notice_cooldowns(expires_at);
 
 -- Create initial checkpoint
 INSERT INTO checkpoints (summary) VALUES ('initial');
--- a/skills/comm-bridge/scripts/__tests__/c4-db-control.test.js
+++ b/skills/comm-bridge/scripts/__tests__/c4-db-control.test.js
@@ -25,6 +25,7 @@ else process.env.ZYLOS_DIR = ORIG_ZYLOS_DIR;
 function resetTables() {
   db.exec('DELETE FROM control_queue');
   db.exec('DELETE FROM conversations');
+  db.exec('DELETE FROM status_notice_cooldowns');
   // Reset autoincrement counters
   db.exec("DELETE FROM sqlite_sequence WHERE name IN ('control_queue', 'conversations')");
 }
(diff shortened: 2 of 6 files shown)


zylos0t:
CLEAN / LGTM.

I reviewed this against #603 using the review-thinking gates, with focus on schema/migration, atomic reserve/suppress, send-failure rollback, suppressed audit semantics, endpoint normalization, and the new concurrency test. I did not find a blocker or a required change.

Evidence path:

- Schema and migration are in the right ownership boundary. `init-db.sql` now defines `conversations.delivery_action` and `status_notice_cooldowns`, and `c4-db.js` runs schema ensures from `getDb()` so existing `c4.db` files are upgraded in place rather than relying only on fresh init. I also smoke-tested an old-style DB without those objects: opening it through the PR code added `delivery_action`, created `status_notice_cooldowns`, preserved the existing conversation row, and allowed reserve/repeat to return reserve then suppress.
- Reserve/suppress is atomic at the SQLite boundary. `reserveStatusNoticeCooldown()` runs in a `better-sqlite3` transaction, prunes expired rows, then uses `INSERT OR IGNORE` against `cooldown_key` as the primary key. Combined with the existing WAL and `busy_timeout`, this composes with the existing comm-bridge DB concurrency mechanism instead of recreating file locks.
- Send failure cleanup preserves the key #599 behavior. `c4-receive.js` reserves before notifying, and only calls `clearStatusNoticeCooldownReservation()` on send failure when it owns a finite `reservedAt`. The clear path deletes by `cooldown_key + last_notified_at`, so it does not wipe out a later reservation from another process.
- Suppressed audit semantics are clearer than #599. Suppressed incoming records remain `status = 'delivered'` so they are not retried by the queue, while `delivery_action = 'suppressed'` makes the actual action queryable. The tests assert both the content marker and `delivery_action` for fallback and router paths.
- Endpoint normalization now has the missing decision path. The code comment explains why volatile `msg` / `req` / `parent` dimensions are stripped while stable root/thread dimensions are preserved, and the tests cover different `parent/msg` values sharing one normalized cooldown row.
- The concurrency test is materially better than the previous sleep-window test. The send helper writes a started marker and waits for an explicit release file, so the second receive is launched after the first has reached the send barrier. The assertions verify one delivered, one suppressed, one outbound send, two inbound records, and one suppressed `delivery_action`.

Verification I ran:

- `git diff --check origin/main...HEAD`
- `node --test skills/comm-bridge/scripts/__tests__/c4-receive.test.js` - 26/26 pass
- `node --test skills/comm-bridge/scripts/__tests__/c4-db-control.test.js` - 49/49 pass
- `npm test` - Jest 79/79 pass, Node 462/462 pass
- Manual migration smoke test from a pre-PR DB shape: pass

Validation environment note: the temporary review worktree needed component dependencies installed under `skills/comm-bridge` for `better-sqlite3`; I treated that as local setup, not a PR finding.

zylos0t (approved):
Approved / LGTM, no blockers.

PR #609, opened 2026-06-03
fix: pin PM2 services to managed Node path

jinglever:
## Summary
- run PM2 service/restart commands with Zylos managed PATH instead of ambient shell PATH
- inject the resolved managed Node interpreter into core JS services and component services loaded by the core ecosystem
- add PM2 env coverage so SYSTEM_PATH stays ahead of Homebrew or other ambient Node paths

## Validation
- node --check templates/pm2/ecosystem.config.cjs
- node --test cli/lib/__tests__/pm2.test.js cli/lib/__tests__/pm2-env.test.js cli/lib/__tests__/service-restart.test.js
- node --test cli/lib/__tests__/upgrade-restart.test.js cli/lib/__tests__/pm2.test.js cli/lib/__tests__/pm2-env.test.js
- npm test: Jest passed, Node suite has one unrelated existing environment failure in cli/lib/__tests__/fs-utils.test.js TMPDIR fallback

## Live recovery
- Dashboard PM2 service restored under Node 24.14.0
- direct dashboard API smoke passed
- local Caddy /dashboard/, /dashboard/api/health, and asset route returned 200

### DIFF
--- a/cli/commands/init.js
+++ b/cli/commands/init.js
@@ -40,6 +40,7 @@ import {
   saveCodexBaseUrlToEnv,
   writeCodexConfig,
 } from '../lib/runtime-setup.js';
+import { buildManagedPm2Env } from '../lib/pm2-env.js';
 
 // Source directories (shipped with zylos package)
 const PACKAGE_ROOT = path.join(import.meta.dirname, '..', '..');
@@ -596,7 +597,7 @@ async function guideBypassAcceptance() {
   console.log(`\n${heading('Setting up Claude Code...')}`);
 
   // Stop activity-monitor to prevent restart loop
-  try { execSync('pm2 stop activity-monitor', { stdio: 'pipe' }); } catch {}
+  try { execSync('pm2 stop activity-monitor', { stdio: 'pipe', env: buildManagedPm2Env() }); } catch {}
 
   // Kill existing session if stuck
   try { execSync(`tmux kill-session -t ${CLAUDE_SESSION} 2>/dev/null`, { stdio: 'pipe' }); } catch {}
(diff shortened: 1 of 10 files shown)


jinglever (commented):
Findings:

1. P1 `templates/pm2/ecosystem.config.cjs:159-162` injects the managed Node interpreter into every component-owned ecosystem app that does not already set `interpreter`. That fixes JS services, but it changes the meaning of a component's own PM2 config and breaks any non-JS/binary app that intentionally relied on PM2's default interpreter resolution. For example, a component app `{ name: 'zylos-bincomp', script: '/usr/bin/true' }` is transformed into `{ script: '/usr/bin/true', interpreter: '<managed node>' }`, so PM2 will try to run `node /usr/bin/true`. This violates the component-owned ecosystem boundary: core should add the managed `PATH`, but should only pin Node for scripts it can prove are Node scripts, or require components to opt in/declare an interpreter. The fallback generated from `SKILL.md` has the same concern at `templates/pm2/ecosystem.config.cjs:184-188` if `service.entry` is not constrained to JS.

2. P2 The PR still leaves some upgrade/self-upgrade PM2 calls on the ambient shell `PATH`, so the managed-PATH guarantee is not complete for the public upgrade paths. `cli/lib/upgrade.js:359` and `cli/lib/upgrade.js:375` call `pm2 jlist` / `pm2 stop` directly, the step-7 recovery path calls `pm2 delete` at `cli/lib/upgrade.js:526`, and self-upgrade verification/save calls `pm2 jlist` / `pm2 save` without `buildManagedPm2Env()` at `cli/lib/self-upgrade.js:1243` and `cli/lib/self-upgrade.js:1263`. In an environment where ambient `PATH` lacks PM2 or resolves to a different global PM2 than `SYSTEM_PATH`, restart may use the managed helper but verification/save can still fail or operate through the wrong PM2 CLI. Since this PR's mechanism is to remove ambient PATH from PM2 orchestration, these remaining direct calls should use the same managed env/helper path.

Validation I ran:
- `node --check templates/pm2/ecosystem.config.cjs`
- `node --test cli/lib/__tests__/pm2.test.js cli/lib/__tests__/pm2-env.test.js cli/lib/__tests__/service-restart.test.js cli/lib/__tests__/upgrade-restart.test.js`
- `node --test cli/lib/__tests__/self-upgrade.test.js`
- `git diff --check origin/main...HEAD`

Verdict: changes still needed. The PR is pointed at the right mechanism, but it currently over-applies Node interpreter policy to component-owned PM2 apps and still has a few ambient PM2 calls in upgrade paths.

PR #633, opened 2026-06-12
fix: auto-retry GitHub API calls on rate limiting in zylos add/upgrade (#632)

### LINKED ISSUE #632: zylos add / zylos upgrade: auto-retry GitHub API calls on rate limiting (up to 2 retries with backoff)
## Problem

`zylos add` and `zylos upgrade` hit GitHub over several paths:

- `cli/lib/github.js` — `fetchRawFile()` (raw.githubusercontent.com → api.github.com/contents fallback), `fetchTagsJsonSync()/fetchTagsJsonAsync()` (api.github.com/repos/:repo/tags)
- `cli/lib/download.js` — `curlDownload()` (github.com archive → api.github.com/tarball fallback)

All of these are single-shot `curl -fsSL` calls. When GitHub rate-limits the request (HTTP 403 primary limit — 60 req/h unauthenticated per IP — or HTTP 429 secondary limit), the command fails immediately and the failure surfaces to the user as a generic download/fetch error. Customer instances are most exposed: often unauthenticated and on shared egress IPs, exactly the conditions where the per-IP 60/h quota is exhausted by neighbors. A real onboarding case surfaced this as user-visible friction.

## Proposal

When a GitHub call fails with a rate-limit signature (HTTP 403/429), automatically retry **up to 2 times** with increasing wait intervals, with a visible "rate limited, retrying in Ns…" notice so the pause is explained. Non-rate-limit failures keep failing fast — no behavior change.

Implementation sketch:
- Shared `withRateLimitRetry()` helper (sync + async flavors) in `cli/lib/github.js`, wrapping each whole operation (the existing public→authenticated fallback chains stay intact inside one attempt, so a rate-limited public call still falls through to the authenticated endpoint *before* any sleep)
- Rate-limit detection from the curl failure (`The requested URL returned error: 403/429`)
- Backoff intervals on the order of 15s / 45s — long enough for secondary limits to clear, bounded enough for a CLI (worst case adds ~60s before final failure)

Out of scope: the one-off caddy release lookup in `cli/commands/init.js` (init-time, different failure handling).

zylos01:
## Summary
- New `withRateLimitRetrySync/Async` helpers in `cli/lib/github.js`: when a GitHub call fails with a rate-limit signature (curl-reported HTTP 403/429), retry up to 2 times with 15s/45s waits and a stderr notice (stderr so `zylos upgrade --json` stdout stays clean). Non-rate-limit failures rethrow immediately — no behavior change.
- Wired into every GitHub path `zylos add`/`zylos upgrade` uses: `fetchRawFile` (SKILL.md fetch), `fetchLatestTag`/`fetchLatestTagAsync` (tags listing), and `curlDownload` (tarball download).
- The retry wraps each operation's **whole** public→authenticated fallback chain, so a rate-limited public call still falls through to the authenticated endpoint *before* any sleep (a token-holding instance never waits needlessly).
- Delays overridable via `ZYLOS_GH_RETRY_DELAY_MS` (comma-separated; empty disables) for tests and ops tuning.

## Notes for review
- Detection is a stderr heuristic (`returned error: 403|429`) because `curl -f` discards the response body/headers. A genuine 403 permission denial would also retry — bounded cost (~60s worst case) on an already-failing path; called out in code comments.
- Sync sleep uses `Atomics.wait` (no child process, no event-loop spin).

## Verification
- 16 new tests in `cli/lib/__tests__/github-rate-limit.test.js`: detection matrix (403/429/404/500/network/digits-in-URL), sync+async retry semantics (success-first, recover-on-retry, exhaust-after-2, fail-fast on non-rate-limit, env disable), plus end-to-end `fetchRawFile` wiring through a fake `curl` shim on PATH (recovers at call 3; non-rate-limit fails fast at 2 calls).
- Full suite: Jest 82/82, node tests 487/487.

Closes #632

### DIFF
--- a/cli/lib/__tests__/github-rate-limit.test.js
+++ b/cli/lib/__tests__/github-rate-limit.test.js
@@ -1,0 +1,191 @@
+import assert from 'node:assert/strict';
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+import { afterEach, beforeEach, describe, it } from 'node:test';
+
+const origEnv = {};
+for (const key of ['ZYLOS_GH_RETRY_DELAY_MS', 'GITHUB_TOKEN', 'GH_TOKEN', 'PATH']) {
+  origEnv[key] = process.env[key];
+}
+const tmpDirs = [];
+
+const { isRateLimitError, withRateLimitRetrySync, withRateLimitRetryAsync, fetchRawFile } =
+  await import('../github.js');
+
+beforeEach(() => {
+  // Fast retries so tests don't sleep for real
+  process.env.ZYLOS_GH_RETRY_DELAY_MS = '10,10';
+});
+
+afterEach(() => {
+  for (const [key, value] of Object.entries(origEnv)) {
+    if (value === undefined) {
+      delete process.env[key];
+    } else {
+      process.env[key] = value;
+    }
+  }
+  while (tmpDirs.length > 0) {
+    fs.rmSync(tmpDirs.pop(), { recursive: true, force: true });
+  }
+});
+
+function curlError(status) {
+  const err = new Error('Command failed: curl -fsSL https://api.github.com/repos/org/repo/tags');
+  err.stderr = `curl: (22) The requested URL returned error: ${status}`;
+  return err;
+}
+
+describe('isRateLimitError', () => {
+  it('matches curl 403 failures', () => {
+    assert.equal(isRateLimitError(curlError('403')), true);
+  });
+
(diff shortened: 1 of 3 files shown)


jinglever (approved):
Scoped review CLEAN for d2f67de. I checked the retry detection, operation-level wrapping, sync/async retry semantics, stderr-only notices, and the 15s/45s bounded backoff trade-off. Targeted new test passes locally; Jest full suite passes locally. Node full suite has one unrelated fs-utils TMPDIR fallback failure that reproduces on origin/main in this environment.

Issue #635, opened 2026-06-15
runtime claude should support write-only credential materialization


nicholas-1vbw:
## Summary

`zylos runtime claude` should support an explicit write-only credential materialization path for Claude runtime configuration.

Today, downstream headless prepare / runtime-switch orchestration cannot safely delegate Claude credential persistence to the supported `zylos runtime` command shape because the Claude path is coupled to live auth validation. That is different from the Codex prepare shape, where callers can persist the API key/base URL through the runtime command before the agent is started.

## Problem

For Codex, downstream prepare flows can use the runtime command as the canonical persistence path:

```sh
zylos runtime codex --save-apikey <key> [--save-base-url <url>]
```

For Claude, downstream orchestration needs the same kind of supported persistence boundary, but a validated Claude runtime path is not safe in prepare-time or runtime-switch contexts. The credentials may be:

- Claude-compatible gateway credentials with non-Anthropic key formats
- mock/test credentials pointed at a controlled base URL
- user-provided values that must be materialized before the runtime is started and validated in its final environment

If the save path performs live validation first, a deployment can fail during prepare even though the final runtime configuration is intentionally valid for the target environment. Downstream callers are then forced to manually write files such as `~/zylos/.env` and Claude settings instead of using one supported `zylos runtime` interface.

This also blocks a clean `switch_runtime` capability in orchestrators: switching from Codex to Claude may require writing Claude env/config first, then selecting the Claude runtime. A command that validates before materializing the Claude credentials cannot support that ordering reliably.

## Expected behavior

Zylos should provide an explicit, documented way to materialize Claude runtime credentials without live auth validation. The exact CLI shape is up to the implementation, for example:

```sh
zylos runtime claude --save-apikey <key> [--save-base-url <url>] --no-validate
```

or a similarly explicit write-only/materialize mode.

The behavior should:

- write/update the same Claude runtime config files that the supported validated path uses
- persist both API key and base URL when provided
- preserve unrelated existing configuration
- avoid contacting the public Anthropic API or otherwise requiring live auth during materialization
- avoid starting the runtime or requiring the runtime to already be healthy
- keep existing interactive/default validation behavior unchanged unless the explicit write-only mode is requested

## Non-goals

- Do not remove auth validation from normal `zylos init` or default `zylos runtime claude` flows.
- Do not silently accept invalid credentials in interactive user flows.
- Do not add another downstream-specific file format; this should use the same canonical Claude config/update semantics as zylos-core.

[...]

PR #641, opened 2026-06-16
fix: runtime auth checks key off explicit signals, not exit codes (Claude + Codex)

zylos-luna:
## Problem

Both runtimes' auth checks decided authentication from a **process exit code**, which is the wrong signal:

### Claude (init)
`isClaudeAuthenticated()` returned `spawnSync('claude', ['auth','status']).status === 0`. The exit code conflates "not logged in" with crashes, timeouts, and cross-version subcommand drift, while the JSON body's explicit `loggedIn` field was piped and discarded.

### Codex (init **and** adapter) — security-relevant
`codex login status` exits **0 in both states** — "Logged in using …" *and* "Not logged in". So keying off the exit code reported a **logged-out OAuth/ChatGPT session as authenticated** (a false-pass). This affected:
- `isCodexAuthenticated()` (init onboarding), and
- `CodexAdapter.checkAuth()` — the gate used by **runtime-switch** and the **activity-monitor health probe**.

The adapter case is the dangerous one: a runtime switch into an unauthenticated Codex would **pass** the auth check and proceed, leaving the agent unreachable over IM — exactly the failure the switch's fail-closed guard exists to prevent (relevant to the #635 design discussion).

## Fix

Key off explicit signals; fail closed on anything ambiguous. Pure parsers live in a new dependency-free `cli/lib/auth-parsers.js`:

- `parseClaudeAuthStatus(stdout)` → `JSON.parse(stdout).loggedIn === true`. `claude auth status --json` passed explicitly so a future default flip to `--text` can't silently break it.
- `parseCodexLoginStatus(text)` → matches a line starting with "Logged in", excludes "Not logged in". The status line is emitted on **stderr** (stdout is empty), so callers pass combined stdout+stderr; the per-line regex ignores any leading warning line.

Extracting the parsers into a leaf module also avoids dragging `runtime-setup.js`'s `child_process` import into adapter test graphs that mock `node:child_process` (which otherwise broke `runtime-launch.test.js`).

Neither check makes a network call — Claude verified via strace (no outbound `connect()`); Codex reads local credential state.

## Tests
- Unit cases for both parsers (logged-in / not-logged-in / missing / truthy-not-true / substring guard / warning-prefixed / empty / undefined).
- Full node suite **506/506** + jest **111/111**.
- Live-verified end to end: real logged-in → authed; isolated-HOME logged-out → not authed — for **both** the init helper and `CodexAdapter.checkAuth()` (the latter now correctly returns `{ok:false, reason:'not_logged_in'}` where it previously false-passed).

## Scope
Decoupled from #635 code-wise, but the Codex adapter fix is a **precondition** for #635's fail-closed guarantee to actually hold for OAuth/ChatGPT Codex. apikey-mode Codex was never affected (it reads auth.json + does a live HTTP probe).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - **Claude init auth check**: `isClaudeAuthenticated()` now keys off the explicit `loggedIn` field from `claude auth status --json` instead of the process exit code. The exit code conflated "not logged in" with crashes, timeouts, and cross-version subcommand drift; the new check reads the unambiguous field and treats any unparseable payload as not authenticated. Extracted a pure, unit-tested `parseClaudeAuthStatus()` helper. Local-only — no network call. (#641)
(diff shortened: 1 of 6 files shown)


jinglever (approved):
R2 CLEAN. Reviewed latest head 999ea49; R1 P3 adapter-level Codex auth false-pass regression test is resolved. Verified locally: git diff --check, runtime-setup.test.js, and codex.test.js all pass.

PR #642, opened 2026-06-16
fix: unify runtime checkAuth into explicit tristate + add --no-validate (#640)

### LINKED ISSUE #640: Unify runtime checkAuth into explicit tristate (success/failure/uncertain) + add --no-validate to runtime switch
## Summary

Refactor runtime `checkAuth()` into an explicit, runtime-unified **tristate** (`success` / `failure` / `uncertain`), move the "is uncertain acceptable?" decision out of the probe and into each caller, make the `zylos runtime` switch gate strict (only confirmed success passes), and add a `--no-validate` flag to skip the switch-time auth probe.

This issue grew out of the design discussion on #635. #635 framed the need as a "write-only credential materialization path" for `zylos runtime claude`. On deeper review, `zylos runtime <name>` is fundamentally a *switch + restart* command (any `--save-*` flag bypasses the already-on-target early-return and triggers a full switch, config update, instruction rebuild, and PM2 restart), and the real blocker is the coupled live auth probe. So the correct shape is not a separate materialize path but: (a) an explicit `--no-validate` to skip the probe, and (b) cleaning up `checkAuth()`'s semantics so the probe reports honestly and each caller decides policy.

## Problem

1. **`checkAuth()` collapses "uncertain" into `ok: true`.** Today it returns `{ ok, reason }` where network errors / timeouts / 429 / 5xx (Codex) and `rate_limit_error` / `api_error` / network error codes (Claude) all return `ok: true`. This bakes a policy decision ("uncertain is acceptable") into the probe itself. An uncertain result may well be a real failure — the probe should report it as uncertain and let the caller decide.

2. **The switch gate cannot distinguish confirmed-success from uncertain.** `cli/commands/runtime.js:258` runs `adapter.checkAuth()` and blocks the switch only on `ok === false`. Because uncertain is `ok: true`, an unverifiable result silently passes the gate.

3. **Claude `checkAuth()` ignores `ANTHROPIC_BASE_URL`.** `cli/lib/runtime/claude.js` injects only `ANTHROPIC_API_KEY` / `CLAUDE_CODE_OAUTH_TOKEN` into the probe env, while `launch()` injects the base URL too. Compatible-gateway / proxy users get a probe sent to the default Anthropic endpoint, producing false failures. (Codex's `checkAuth()` is already base-URL aware via `getCodexApiBaseUrl()`. This is the Claude-side analogue of #434.)

4. **No way to skip validation for headless / prepare / mock / gateway contexts.** There is no `--no-validate`, so orchestration that intends a valid-in-final-environment switch is blocked at prepare time by a probe that cannot succeed yet.

### 1. `checkAuth()` → explicit tristate, unified across runtimes

Return `{ status: 'success' | 'failure' | 'uncertain', reason }` (drop `ok`). Both adapters map their raw signals into the same three buckets:

| Bucket | Claude (`claude -p ping`) | Codex (HTTP probe / `codex login status`) |
|---|---|---|
| **success** | probe ok (`cli_probe`) | `http_probe_200`, `codex_login_status` |
| **failure** | `authentication_error`, "Not logged in" | `401`, not logged in, apikey-mode-but-no-key |
| **uncertain** | `rate_limit_error`, `api_error`, `ETIMEDOUT`/`ECONNREFUSED`/`ENOTFOUND`, killed | 429, 5xx, other non-200, network error / timeout |

The probe no longer decides whether uncertain "passes" — it only reports.

### 2. Adapt every caller to the tristate (explicit per-site policy)

`checkAuth()` has 6 call sites; each declares its own policy:

| Caller | success | failure | uncertain |
|---|---|---|---|
| `cli/commands/runtime.js:258` (switch gate) | proceed (switch) | block (exit 2) | **block (exit 2)** |
| `skills/activity-monitor/scripts/health-engine.js:520` (set auth_failed) | no-op | set `auth_failed` | **tolerate (no-op)** |
| `skills/activity-monitor/scripts/health-engine.js:576` (recovery check) | clear | stay failed | **clear (benefit of the doubt)** |
| `cli/commands/service.js:92` (status display) | authenticated | NOT AUTHENTICATED | **show "unverified"** |
| `cli/commands/doctor.js:297/302` (diagnostics) | ✓ | ✗ | **unverified** |
| `skills/activity-monitor/scripts/adapters/runtime-components.js:45` (fallback shim) | — | — | fallback return updated to tristate |

**Invariant (safety-critical):** the health-engine must continue to treat `uncertain` as tolerated — only a confirmed `failure` may flip health to `auth_failed`. This must be locked by tests to avoid false restarts on transient network blips.

### 3. Strict switch gate + `--no-validate`

- Switch gate passes only when `status === 'success'`. Both `uncertain` and `failure` abort the switch (exit 2). Default-block is retained intentionally: an unattended agent that switches into an unreachable runtime loses IM access and cannot self-recover.
- `--no-validate` is an independent flag that skips the switch-time `checkAuth()` entirely. It does **not** require `--save-apikey` / `--save-setup-token` (orthogonal concern), and is symmetric for both runtimes. The existing already-on-target no-op early-return is preserved, so bare `--no-validate` only does work on a genuine runtime change.

### 4. Fix Claude `checkAuth()` base URL

Inject `ANTHROPIC_BASE_URL` from `~/zylos/.env` into the probe env, matching `launch()`. This becomes more important under the strict gate (a gateway probe hitting the wrong endpoint would now hard-block the switch).

## Out of scope (follow-up)

- **Auto-rollback safety net:** when a switch made with `--no-validate` results in a runtime that cannot establish a session, activity-monitor should record the previous runtime and roll back. This touches the health-engine and warrants its own PR/review; tracked separately.

## Non-goals

- Do not remove auth validation from the default (non-`--no-validate`) interactive switch or `zylos init`.
- Do not change the health-engine's tolerance of uncertain results (must remain tolerated).

## Acceptance criteria

- `checkAuth()` returns `{ status, reason }` with unified semantics across Claude and Codex; no `ok` field.
- All 6 call sites updated to explicit per-status policy per the table above.
- Switch gate passes only on `status === 'success'`; uncertain and failure both abort (exit 2).
- `--no-validate` skips the switch-time probe, works standalone, symmetric for both runtimes.
- Claude `checkAuth()` injects `ANTHROPIC_BASE_URL`.
- Tests lock: switch gate blocks on uncertain; `--no-validate` does not call `checkAuth()`; health-engine does NOT set `auth_failed` on uncertain; tristate mapping for both adapters.

## Related

- #635 — original write-only materialization framing that surfaced this (this issue refines that approach).
- #434 — Codex-side precedent (base-URL-aware auth check).
- #399 — runtime-specific base URL command shape.

zylos-luna:
Closes #640.

## What & why

Refactors runtime `checkAuth()` from `{ ok, reason }` into an explicit **tristate** `{ status: 'success' | 'failure' | 'uncertain', reason }`, pushes the "is uncertain acceptable?" decision out of the probe and into each caller, makes the `zylos runtime` switch gate strict, and adds a standalone `--no-validate` flag.

The core problem: today `uncertain` (network error / 429 / 5xx / timeout) is silently folded into `ok: true`. That lets an unverifiable result pass the switch gate, while the health-engine *needs* to tolerate transient blips — two opposite policies tangled in the probe. The tristate makes the probe report honestly and each caller decide.

## Changes

- **`checkAuth()` → tristate**, unified across Claude + Codex; `ok` field removed (`base.js`, `claude.js`, `codex.js`).
- **Claude probe injects `ANTHROPIC_BASE_URL`** (was only injecting key/token; gateway/proxy users got false failures). Verified by an env-capture test.
- **Codex**: new `classifyCodexLoginStatus()` 3-way classifier in `auth-parsers.js` (Logged in→success / Not logged in→failure / unparseable·execFile-threw→uncertain). `parseCodexLoginStatus()` kept as a thin boolean view for the init flow.
- **6 caller sites** migrated to explicit per-status policy: switch gate (`runtime.js`), status (`service.js`), doctor (`doctor.js`), health-engine set-auth_failed / recovery / wrapper, runtime-components shim.
- **Strict switch gate**: only `status === 'success'` proceeds; `uncertain` and `failure` both exit 2 (distinct messages). Extracted to a testable `checkRuntimeAuthGate()`.
- **`--no-validate`**: standalone flag (no `--save-*` pairing required), symmetric for both runtimes, skips only the switch-time probe; already-on-target no-op early-return preserved.

**Safety invariant (locked by tests):** the health-engine treats `uncertain` as tolerated — only a confirmed `failure` flips to `auth_failed` (no false restarts on transient network blips). The recovery probe only restarts on `success`.

## Scope note (off-scope fix, transparent disposition)

The second commit, `fix(fs-utils): copyTree self-copy detection under symlinked tmp`, is a correct but **off-scope** 4-line fix unrelated to the auth change. It corrects self-copy detection on macOS (`/var` → `/private/var`), which otherwise fails the full Node test suite locally. Kept in this PR (rather than a separate 4-line PR) but isolated in its own commit; flagged here per reviewer request.

## Tests

- Full suite green: **Jest 111/111, Node 523/523**.
- New coverage: Claude success/failure/uncertain + `ANTHROPIC_BASE_URL` injection; Codex 200/401/429/5xx/network + login-status logged-in/not-logged-in/unparseable/execFile-threw→uncertain; gate `--no-validate` skip (checkAuth 0 calls) + success-only-pass / failure·uncertain→exit 2; health-engine regression locks (uncertain never sets auth_failed; recovery uncertain stays auth_failed + killTmuxSession 0).

## Review

Independent dual review (zylos01 PM/QA + zylos0t): both **CLEAN, no P1/P2**. Three P3s addressed (commit split, execFile→uncertain test, JSDoc).

## Follow-up (separate issue)

Acceptance surfaced a pre-existing probe-robustness gap (not a regression): `claude -p ping` inherits the user-level model setting, so an inaccessible model (e.g. a leftover Fable 5 override) makes the auth probe report failure despite valid credentials. A focused follow-up will pin the probe to a stable always-available model and reconsider whether unknown-nonzero outcomes should map to `uncertain`.


jinglever (approved):
LGTM — concurs with prior dev-workflow review (tristate refactor + safety invariants test-locked). Approving to satisfy non-author approval rule.

PR #656, opened 2026-06-21
feat: Composio MCP integration (Tool Router, JIT tool-scoping)

FelixLin6:
## Summary

Lets any Zylos instance opt into **Composio** (MCP tool-aggregator, ~1,000+ apps with JIT tool-scoping via the Tool Router) by setting `COMPOSIO_API_KEY` in `.env`. One Composio account per instance; the MCP URL is generated/reused automatically during init / runtime setup.

Six integration pieces: `templates/.mcp.json`, the sync step, `enabledMcpjsonServers` in settings, the Codex `config.toml` block, blank `.env` vars, and runtime-switch reapply. Works for both the Claude Code and Codex runtimes.

## Security posture

- **Remote code execution disabled at the source.** The Tool Router session is minted with top-level `workbench: { enable: false }`, which removes `COMPOSIO_REMOTE_BASH_TOOL` and `COMPOSIO_REMOTE_WORKBENCH` from the session entirely (verified live: session tools drop 6 → 4). The runtime allow/deny list is retained as defense-in-depth. A session that still exposes those tools is rejected.
- Session mint requires `user_id` (v3 API — a `{}` body returns 400 and silently broke the mint before this change).
- API key is never placed in argv — it is passed to `curl` via `--config` on stdin.
- Generated `.env` / config writes are atomic (tmp + rename, mode 0600).
- Idempotent re-runs never clobber or silently adopt a user's pre-existing, unmarked `composio` entry (Claude `.mcp.json` and Codex `config.toml`); cleanup only removes zylos-marked entries.

## Prod-readiness review

Partitioned review found 1 HIGH + 2 MEDIUM + 3 LOW; all reconciled and patched:
- **H1** — bash/workbench bypass via `MULTI_EXECUTE` → closed via `workbench:{enable:false}` (+ required `user_id`).
- **M1** — Codex config cleanup is marker-only (never clobbers user entries).
- **M2** — atomic `.env` write (tmp+rename, 0600).
- **L1** — API key kept off the command line.
- **L3** — deny-by-default template posture (no permissive `defaultMode`; only the 4 meta tools allowed; bash/workbench denied).

## Verification (independent VM pass on `3c5bac3`)

- H1 confirmed **live** against a real Composio account: 6→4 tools, no BASH/WORKBENCH; empty `{}` body → 400.
- Node test suite: **550/550** pass; Composio files (composio-mcp / runtime-setup / sync-settings-hooks) **91/91**.
- `git diff --check` clean; syntax checks pass.

> ⚠️ **Cascade risk:** zylos-core upgrades cascade to ALL instances. Merge only after the independent reviewer pass + final VM sign-off.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,9 @@ node_modules/
 *.log
 .DS_Store
 .env
+/.mcp.json
+/.mcp.zylos.json
+/.codex/composio.zylos.json
 *.db
 *.db-shm
 *.db-wal
--- a/cli/commands/init.js
+++ b/cli/commands/init.js
@@ -20,6 +20,7 @@ import { commandExists } from '../lib/shell-utils.js';
 import { getActiveAdapter } from '../lib/runtime/index.js';
 import { buildInstructionFile } from '../lib/runtime/instruction-builder.js';
 import { deployManifestTemplate } from '../lib/runtime/tmux-env.js';
+import { syncClaudeComposioMcpJson } from '../lib/composio-mcp.js';
 import { runMigrations } from '../lib/migrate.js';
 import {
   installGlobalPackage,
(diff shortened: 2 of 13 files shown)


zylos0t (approved):
Approved after R2 review at b4ffb76f2e74c45959b45976a3709e6ecd44bfff. Prior Composio MCP persistence/security findings are closed; verification passed: git diff --check, npm test (Jest 111/111, Node 553/553), and manual repros for same-shape unmarked Claude config preservation, malformed Codex TOML preservation, and Codex config mode 600.

PR #662, opened 2026-06-23
fix: guard opus[1m] model backfill against incompatible threshold

### LINKED ISSUE #661: upgrade: opus[1m] model backfill should guard against incompatible new_session_threshold
## Problem

When upgrading older instances, `syncTemplateModelSetting()` can backfill `opus[1m]` without ensuring the `new_session_threshold` is compatible, resulting in a high-cost configuration.

**Confirmed reproduction path:**

1. `syncTemplateModelSetting()` checks if installed `.claude/settings.json` is missing `model` — if so, backfills from template → `opus[1m]`
2. `persistInstalledSettingsAndSyncCoupledThreshold()` then calls `syncModelCoupledNewSessionThreshold()`
3. `syncModelCoupledNewSessionThreshold()` finds `.zylos/config.json` already has `new_session_threshold` (e.g. the old default `70`) → returns `{ changed: false, reason: 'threshold_already_set' }`, does **not** adjust to `30`
4. Result: old instance ends up with `opus[1m]` + 70% threshold — effectively using ~700K context per session, which is the exact high-consumption pattern reported in #659

**Function-level repro:**
- Installed settings: `{}`
- Template model: `opus[1m]`
- Config: `{ new_session_threshold: 70 }`
- After upgrade: settings → `{ model: 'opus[1m]' }`, threshold sync → `{ changed: false, reason: 'threshold_already_set' }`

The `self-upgrade` fallback path has the same gap: `generateMigrationHints()` / `applyMigrationHints()` generate and apply `model_backfill` but have no threshold preflight/guard.

## Relevant code

- `cli/lib/sync-settings-hooks.js`: `syncTemplateModelSetting()`, `syncModelCoupledNewSessionThreshold()`, `persistInstalledSettingsAndSyncCoupledThreshold()`

## Proposed fix

Make the `opus[1m]` model backfill and threshold migration an atomic/guarded operation. Two possible strategies:

1. **Guard the backfill (preferred):** If `new_session_threshold` already exists and cannot be safely migrated to ≤30, do **not** auto-backfill `opus[1m]` — leave the model unset or fall back to `opus` (200K). Rationale: there is currently no provenance to distinguish "user explicitly set 70" from "system old default 70", so we cannot safely assume the threshold is safe to overwrite.

2. **Coupled migration:** If the old default 70 is confirmed safe to migrate, change it to 30 in the same backfill operation, with explicit test coverage documenting this as intentional migration.

Strategy 1 is safer given the lack of provenance tracking.

## Related

- #659 (original report of high token consumption with opus[1m])
- Analysis by zylos0t confirming the upgrade path gap

zylos-luna:
## Summary

- When upgrading older instances, `syncTemplateModelSetting()` backfills `opus[1m]` without checking whether `new_session_threshold` is compatible — an instance with threshold=70 ends up with opus[1m]+70% (~700K context/session)
- Before backfilling a `[1m]` model, now checks if `new_session_threshold` > 30; if so, strips `[1m]` suffix (e.g. `opus[1m]` → `opus`)
- Applies to both the regular upgrade path (`syncTemplateModelSetting`) and the self-upgrade fallback (`generateMigrationHints`)

## Changes

- `cli/lib/sync-settings-hooks.js`: Added threshold guard to `syncTemplateModelSetting()` — accepts `cfg` param, checks `new_session_threshold` before applying 1M model
- `cli/lib/self-upgrade.js`: Same guard in `generateMigrationHints()` — downgrades the hint value before it's generated
- Tests: 6 new test cases covering guard triggers, passthrough, and edge cases

## Test plan

- [x] `node --test cli/lib/__tests__/sync-settings-hooks.test.js` — 43 pass
- [x] `node --test cli/lib/__tests__/self-upgrade.test.js` — 22 pass
- [x] `npm test` — 532 pass, 0 fail

Closes #661

### DIFF
--- a/cli/lib/__tests__/self-upgrade.test.js
+++ b/cli/lib/__tests__/self-upgrade.test.js
@@ -291,6 +291,28 @@ describe('Claude model migration hints', () => {
     fs.rmSync(tmpDir, { recursive: true, force: true });
   });
 
+  it('downgrades 1m model in hint when threshold is above 30', () => {
+    const tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'zylos-model-guard-'));
+    const templatesDir = path.join(tmpDir, 'templates');
+    const zylosDir = path.join(tmpDir, 'zylos');
+
+    fs.mkdirSync(path.join(templatesDir, '.claude'), { recursive: true });
+    fs.mkdirSync(path.join(zylosDir, '.claude'), { recursive: true });
+    fs.writeFileSync(path.join(templatesDir, '.claude', 'settings.json'), JSON.stringify({ model: 'opus[1m]' }), 'utf8');
+    fs.writeFileSync(path.join(zylosDir, '.claude', 'settings.json'), JSON.stringify({ hooks: {} }), 'utf8');
+
+    const hints = generateMigrationHints(templatesDir, {
+      zylosDir,
+      getConfig: () => ({ new_session_threshold: 70 }),
+    });
+    assert.deepEqual(
+      hints.filter((hint) => hint.type === 'model_backfill'),
+      [{ type: 'model_backfill', value: 'opus' }]
+    );
+
+    fs.rmSync(tmpDir, { recursive: true, force: true });
+  });
+
   it('does not add a model backfill hint when the user already configured model', () => {
     const tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'zylos-model-nohint-'));
     const templatesDir = path.join(tmpDir, 'templates');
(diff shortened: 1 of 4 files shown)


zylos0t (approved):
Reviewed head 5cc7957b881248f868dcbed154e8c84b381f0868. CLEAN: the model backfill guard now strips the [1m] suffix when an existing new_session_threshold is above 30, and the self-upgrade hint path applies the same strategy without mutating threshold. Verified with focused node tests, full npm test, git diff --check, and function-level probes for both main sync and self-upgrade fallback paths.

PR #668, opened 2026-06-25
feat: consolidate SessionStart hooks into single orchestrator (#651)

zylos-luna:
## Summary

- Design document for consolidating 4 SessionStart hook commands into a single orchestrator script
- Addresses fragility (no error isolation, #601 hang bug), 3× duplication across matchers, and unbounded startup time
- Orchestrator runs steps in defined order: memory inject → C4 init (sequential, stdout) → foreground + prompt (parallel, side effects)
- Per-step timeout + total 15s budget ensures startup never hangs

This PR starts with the design doc only. Implementation follows after review.

## Design highlights

- **In-process imports** instead of 4 child process spawns (refactor existing scripts to export callable functions)
- **Phase 1** (sequential): memory inject + C4 init → stdout for Claude context
- **Phase 2** (parallel): foreground session file + control-queue prompt → side effects only
- **Error isolation**: each step wrapped in try/catch + AbortSignal.timeout; failure logged and skipped

## Test plan

- [ ] Review design document (`docs/impl-session-start-orchestrator.md`)
- [ ] Confirm approach for open questions (compact skip prompt? settings migration?)

Fixes #651

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #651, opened 2026-06-17
Consolidate SessionStart hooks into a single orchestrator script (memory inject + c4 init + monitor)


zylos-luna:
## Problem

`SessionStart` currently runs **four separate hook commands**, duplicated across the `clear` / `compact` / `startup` matchers in `.claude/settings.json`:

1. `zylos-memory/scripts/session-start-inject.js` (memory injection — identity/state/references)
2. `comm-bridge/scripts/c4-session-init.js`
3. `activity-monitor/scripts/session-start-prompt.js`
4. `activity-monitor/scripts/session-foreground.js`

Each is an independent process spawn with its own timeout, sitting in the **critical session-bootstrap path**. This fragmentation is fragile:

- A hang or failure in any one hook can block or partially break startup. At least one of them already does this — see #601 (`session-start-prompt.js` can hang after execution).
- The four-command list is copy-pasted across every matcher, so any change must be made in N places and they drift.
- There is no shared ordering / error-isolation guarantee: if memory injection is supposed to run before the others, nothing enforces it, and a non-memory hook failing can still disrupt the session.

This has caused real problems (memory/bootstrap reliability), so it is the priority item.

## Goal

Consolidate the SessionStart hooks into a **single orchestrator script** that the matcher invokes once, which runs the individual steps in a defined order with per-step error isolation and a bounded total runtime.

## Proposed behavior

- One `command` entry per matcher in `.claude/settings.json` (e.g. `node ~/zylos/.claude/skills/<...>/session-start.js`).
- The orchestrator runs each step (memory inject → c4 init → monitor prompt → foreground) sequentially or with controlled concurrency.
- **Per-step isolation**: a step that throws/times out is logged and skipped; it must not abort the remaining steps.
- **Bounded exit**: the orchestrator force-exits within a total budget so it can never hang the startup path (addresses the class of bug in #601).
- Behavior parity: same net effect as the current four hooks on `clear` / `compact` / `startup`.

## Acceptance criteria

- `.claude/settings.json` has a single SessionStart command per matcher.
- The orchestrator invokes all four current responsibilities with deterministic ordering.
- One step failing/hanging does not prevent the others from running, and the process exits within a bounded time.
- Template (`templates/...settings...`) updated to match so new installs get the consolidated hook.
- Tests cover step isolation (one step throws → others still run) and bounded exit.

## Related

- #601 — `session-start-prompt.js` can hang after execution (one of the four hooks; consolidation should make the whole path defensive).

PR #671, opened 2026-06-28
fix: unify session-start context format + surface failures and null summaries

zylos-luna:
## Background

Howard ran the merged `session-start-orchestrator.js` and noticed the injected context construction was inconsistent: memory sections use `=== XXX ===` (header only, no footer) while C4 sections use `[Recent Conversations]` / `[Last Checkpoint Summary]`. He also asked to confirm the checkpoint-summary extraction — which works, but had a silent-drop edge.

Follow-up to #651 / #668 (the SessionStart orchestrator line). Scope confirmed by Howard ("按建议来").

## What changed

**1. Unified format (core)** — new shared `formatSection(label, content)` in `comm-bridge/scripts/session-format.js`, used by:
- `zylos-memory/session-start-inject.js` (memory sections)
- `comm-bridge/c4-session-init.js` (checkpoint / conversations / action-required)
- `activity-monitor/session-start-orchestrator.js` (failure notices)

Every block is now `=== LABEL ===` … `=== END LABEL ===` (head **and** tail), uppercase labels throughout.

**2. Visible step failures** — `runStep` now writes a visible `=== <STEP> UNAVAILABLE ===` notice to the injected stdout stream when a context-producing step fails or times out, instead of silently dropping the section (the previous behavior is exactly why a failed step read as "this context simply doesn't exist"). Side-effect steps (foreground/prompt) stay silent on stdout as before.

**3. Checkpoint null-summary fallback** — when a checkpoint exists but `summary` is null, emit `=== LAST CHECKPOINT ===\n(no summary — checkpoint #id, ts)` so the block never silently disappears.

## Dependency direction

`session-format.js` lives in comm-bridge (the most foundational of the three skills — zylos-memory already imports comm-bridge's `c4-diagnostic.js`, and the orchestrator already imports from comm-bridge). comm-bridge itself gains no new cross-skill dependency.

## Out of scope

`c4-fetch.js` also emits a `[Last Checkpoint Summary]` line, but that's the Memory-Sync fetch path, not session-start injection — left untouched.

## Tests

- New `session-format.test.js` unit suite (header/footer, trim, empty fallback, multi-line).
- `session-start-orchestrator.test.js`: the 4 tests that asserted silent-drop now assert the visible failure notice; +2 unit tests (writeStdout step emits notice; side-effect step stays silent).
- `c4-session-init-cli.test.js`: updated for the new uppercase labels + footers; new null-summary fallback test.
- Full suite green: **Jest 111/111, Node 587/587**.
- Real-system smoke: memory + C4 both render uniform `=== LABEL ===`/`=== END ===`, null-summary fallback renders `(no summary — checkpoint #N, ts)`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Changed
-- **Unified session-start context format**: memory injection (`session-start-inject.js`) and C4 session init (`c4-session-init.js`) now share a single `formatSection()` helper, emitting consistent `=== LABEL ===` … `=== END LABEL ===` blocks instead of the previous mixed `=== LABEL ===` (header-only) / `[Bracket Label]` styles, so the combined injected context reads uniformly. (#651)
+- **Unified session-start context format**: memory injection (`session-start-inject.js`) and C4 session init (`c4-session-init.js`) now share a single `formatSection()` helper, emitting consistent `=== LABEL ===` … `=== END LABEL ===` blocks instead of the previous mixed `=== LABEL ===` (header-only) / `[Bracket Label]` styles, so the combined injected context reads uniformly. (#671)
 
 ### Fixed
-- **Visible session-start step failures**: the orchestrator now writes a visible `=== <STEP> UNAVAILABLE ===` notice to the injected stdout stream when a context-producing step fails or times out, instead of silently dropping the section. (#651)
-- **Checkpoint null-summary fallback**: the last-checkpoint block now renders a `(no summary — checkpoint #id, ts)` fallback when a checkpoint exists with a null summary, so the block never silently disappears. (#651)
+- **Visible session-start step failures**: the orchestrator now writes a visible `=== <STEP> UNAVAILABLE ===` notice to the injected stdout stream when a context-producing step fails or times out, instead of silently dropping the section. (#671)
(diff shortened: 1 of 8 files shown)


zylos0t (dismissed):
S2 review: CLEAN / APPROVE.\n\nChecked the PR delta against base 773b239. No blocker found.\n\nFocus points:\n- Shared formatSection import direction is acceptable for the core-skill bundle: comm-bridge owns the shared session-start framing helper, and zylos-memory/activity-monitor import it through stable sibling skill paths.\n- runStep only writes visible UNAVAILABLE sections for writeStdout context steps; side-effect steps remain silent on stdout when they fail.\n- checkpoint summary=null now emits a LAST CHECKPOINT fallback block instead of disappearing.\n- Label casing/renaming looks safe: rg found no session-start parser depending on the old [Recent Conversations] / [Last Checkpoint Summary] labels. c4-fetch keeps its separate Memory Sync fetch label, which is out of this output path.\n\nLinux verification:\n- git diff --check 773b239...HEAD: pass\n- targeted node tests (session-format, c4-session-init-cli, session-start-orchestrator): 34/34 pass after installing skill deps\n- npm test: Jest 111/111 pass; Node 587/587 pass\n\nNote: initial targeted c4-session-init test run failed because the isolated review worktree had not installed skill deps (missing better-sqlite3), then passed after running scripts/install-skill-deps.js. Not PR-introduced.

zz-howard (approved):

PR #674, opened 2026-06-29
fix: lower Memory Sync checkpoint threshold 30→15 + single source of truth (#673)

### LINKED ISSUE #673: Lower Memory Sync checkpoint threshold 30→15 and dedupe CHECKPOINT_THRESHOLD to a single source
## Problem

The Memory Sync checkpoint threshold is `CHECKPOINT_THRESHOLD = 30`. It does two jobs:

1. **Session-start injection** (`comm-bridge/scripts/c4-session-init.js`): when unsummarized conversations ≤ threshold, **all** of them are injected verbatim at session start (so up to 30 raw messages). Over threshold → inject only the last `SESSION_INIT_RECENT_COUNT` (6) and emit the Memory Sync `ACTION REQUIRED` block.
2. **Early in-session sync** (`activity-monitor` context monitor): when context usage is in `[round(new_session_threshold×0.8), new_session_threshold)` **and** unsummarized count > threshold, enqueue a background Memory Sync before the session-switch fires.

Two issues:

- **30 is too large for session-start injection.** Session start already injects identity + state + references (all sizable). Up to 30 raw messages (with `reply via:` boilerplate) can add ~10k+ tokens on top, while the resume value lives in the last handful. The invariant that must hold: the agent always sees every message not yet folded into a checkpoint summary — so the fix is to lower the threshold (shorter unsummarized tail), **not** to cap injection independently (that would lose context for messages between the cap and the threshold).
- **The constant is duplicated in three places**, each a hardcoded literal with a "must match c4-config.js" comment — silent-divergence hazard:
  - `skills/comm-bridge/scripts/c4-config.js:57` (canonical export)
  - `skills/activity-monitor/scripts/context-monitor.js:41`
  - `skills/activity-monitor/scripts/monitor.js:568`

## Fix

1. Lower the threshold **30 → 15**. Worst-case session-start injection roughly halves; the no-context-loss invariant still holds. Cost: slightly more frequent Memory Syncs, each a cheap background subagent.
2. **Single source of truth**: keep the definition only in `c4-config.js` and import it in the two activity-monitor files (`context-monitor.js`, `monitor.js`), removing the hardcoded copies. Cross-package sibling imports are already used here (`monitor.js` imports `../../comm-bridge/scripts/tmux-input-state.js`), so this follows existing precedent. The activity-monitor engine (`adapters/runtime-components.js`) is already parameter-driven and needs no change.

## Out of scope

- `SESSION_INIT_RECENT_COUNT` (6) — unchanged.
- Making the threshold config-driven — not needed now; trivial to add later if it needs frequent tuning.

zylos-luna:
## What

- Lower the Memory Sync checkpoint threshold **30 → 15**.
- Eliminate the **three** hardcoded copies of `CHECKPOINT_THRESHOLD` (`c4-config.js`, `context-monitor.js`, `monitor.js`) → single source of truth in `c4-config.js`, imported by the two activity-monitor files.

## Why

`CHECKPOINT_THRESHOLD` does double duty:
1. **Session-start injection** — under threshold, *all* unsummarized messages are injected verbatim (up to 30 raw messages with `reply via:` boilerplate, on top of identity/state/references). 30 is too large.
2. **Early in-session sync** — fires when context is in `[round(new_session_threshold×0.8), new_session_threshold)` and unsummarized > threshold.

Lowering to 15 ~halves worst-case session-start injection while preserving the invariant that the agent always sees every message not yet folded into a checkpoint summary (just a shorter tail). Cost: slightly more frequent Memory Syncs, each a cheap background subagent.

The constant was duplicated with "must match c4-config.js" comments — a silent-divergence hazard. Consolidated to one export; the activity-monitor files import it via the same sibling cross-package path already used for `tmux-input-state.js`. The context-monitor engine (`adapters/runtime-components.js`) is already parameter-driven, no change needed.

## Test

- Jest: 111/111 ✓
- Node: 587/587 ✓
- Existing session-init threshold tests still valid (3 msgs < 15 → no sync; 31 msgs > 15 → sync); stale "is 30" comment updated.

## Notes for reviewer

- `SESSION_INIT_RECENT_COUNT` (6) unchanged.
- Not made config-driven (kept simple); trivial to add later if it needs frequent tuning.

Closes #673

### DIFF
--- a/skills/activity-monitor/scripts/context-monitor.js
+++ b/skills/activity-monitor/scripts/context-monitor.js
@@ -19,6 +19,7 @@ import { execFileSync } from 'child_process';
 import fs from 'fs';
 import path from 'path';
 import os from 'os';
+import { CHECKPOINT_THRESHOLD } from '../../comm-bridge/scripts/c4-config.js';
 
 const ZYLOS_DIR = process.env.ZYLOS_DIR || path.join(os.homedir(), 'zylos');
 const AM_DIR = path.join(ZYLOS_DIR, 'activity-monitor');
@@ -38,7 +39,7 @@ const COOLDOWN_SECONDS = 300;   // Re-trigger after 5 minutes if still above thr
 // completes in the background before new-session fires.
 const MEMORY_SYNC_RATIO = 0.8;
 const MEMORY_SYNC_THRESHOLD = Math.round(RESTART_THRESHOLD * MEMORY_SYNC_RATIO);
-const CHECKPOINT_THRESHOLD = 30;  // must match c4-config.js CHECKPOINT_THRESHOLD
+// CHECKPOINT_THRESHOLD imported from c4-config.js (single source of truth).
 const MEMORY_SYNC_COOLDOWN_SECONDS = 600;  // 10 min — prevent re-inject while sync is running
 
 function readThresholdFromConfig() {
--- a/skills/activity-monitor/scripts/monitor.js
+++ b/skills/activity-monitor/scripts/monitor.js
@@ -147,6 +147,7 @@ import {
 } from './adapters/runtime-components.js';
 import { createActivityMonitorTaskScheduler } from './tasks/activity-monitor-tasks.js';
 import { readTmuxInputState } from '../../comm-bridge/scripts/tmux-input-state.js';
+import { CHECKPOINT_THRESHOLD } from '../../comm-bridge/scripts/c4-config.js';
 // activity-monitor runs as a deployed skill at ~/zylos/.claude/skills/activity-monitor/scripts/.
 // A relative import to cli/lib/runtime/ resolves correctly in the repo (dev) but NOT from
 // the deployed path — the CLI lives in the globally installed zylos npm package.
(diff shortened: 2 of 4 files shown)


jinglever (approved):
R1: CLEAN / APPROVED

Findings: none.

Review notes:
- Confirmed the PR now has a single threshold source at `skills/comm-bridge/scripts/c4-config.js` with `CHECKPOINT_THRESHOLD = 15`.
- Confirmed both early in-session memory-sync paths consume that shared value: Claude statusLine path in `context-monitor.js`, and runtime adapter path via `monitor.js` -> `runtime-components.js`.
- Checked the sibling import boundary. `activity-monitor -> comm-bridge/scripts/c4-config.js` matches the existing sibling import pattern already used for `tmux-input-state.js`; `c4-config.js` adds only module-level path/config derivation and fail-open config read, so the statusLine hot path failure surface is acceptable.
- Confirmed the session-start invariant still holds: when unsummarized count is <= threshold, all unsummarized messages are injected; when it is > threshold, startup shows the checkpoint context plus recent tail and emits the Memory Sync action block. Lowering 30 -> 15 changes trigger timing/tail size pressure, not the recovery mechanism.

Validation run:
- `git diff --check origin/main...HEAD` passed.
- `npm test -- skills/comm-bridge/scripts/__tests__/c4-session-init-cli.test.js` passed; this repo script ran Jest 111/111 and Node 587/587.
- `node --input-type=module -e "await import('./skills/activity-monitor/scripts/context-monitor.js')" </dev/null` passed.
- `c4-config.js` import reports `CHECKPOINT_THRESHOLD=15`, `SESSION_INIT_RECENT_COUNT=6`.

No configured GitHub checks were present on the PR at review time.

PR #675, opened 2026-06-29
fix: migrate Codex bootstrap to native SessionStart hook

### LINKED ISSUE #652: Migrate Codex runtime to native SessionStart hook (parity with Claude; reuse #651 orchestrator)
## Background

Recent Codex releases now support a **SessionStart hook** mechanism (analogous to Claude Code's). Today the Codex runtime bootstraps its session differently from Claude — it does not go through a native session-start hook for memory injection / C4 session init / monitor prompt, so the two runtimes have divergent startup paths.

## Goal

Migrate the Codex runtime's session bootstrap to use Codex's native SessionStart hook, bringing it to parity with the Claude path (ideally reusing the same consolidated orchestrator from #651).

## Why

- **Parity**: one bootstrap model across runtimes is easier to reason about and maintain than two divergent ones.
- **Reliability**: the native hook is the supported integration point; the current Codex-specific bootstrap is more bespoke.
- **Reuse**: once #651 consolidates the Claude SessionStart hooks into a single orchestrator, Codex can point its session-start hook at the same (or a thin runtime-aware wrapper around the same) script.

## Scope / open questions

- Confirm the exact Codex hook config shape and event names in the new version, and which Codex version first supports it (gate on availability).
- Decide whether Codex and Claude share one orchestrator with runtime branching, or each gets a thin adapter.
- Ensure no double-execution (current Codex bootstrap path must be removed/disabled when the hook takes over).

## Priority

Lower urgency than #651 — this is an improvement/consolidation, not an active fault. Best done after (or together with) #651 so both runtimes converge on the single orchestrator.

## Related

- #651 — Consolidate SessionStart hooks into a single orchestrator (do first; Codex can reuse it).
- #311 — multi-runtime support (pluggable agent backend).

jinglever:
## Summary
- install a core-managed Codex `SessionStart` hook that runs `session-start-orchestrator.js` from shared `hooks.json` while preserving other hook groups
- enable `[features] hooks = true` in project and global Codex config, and mirror that in `templates/.codex/config.toml`
- add a launch-time trust backstop that validates hook feature flags, `hooks.json` hash, Codex version, and the exact saved `hooks.state` snapshot before skipping app-server re-trust
- remove the retired text bootstrap prompt path from Codex launch to avoid double startup injection
- preserve the onboarding guard in the native-hook path by skipping startup prompt enqueue while onboarding is pending

## R1 updates
- added fail-closed coverage for `empty_trust_snapshot` when app-server returns success but writes no matching `hooks.state` entries
- removed the unused `trustStateFromHooksList` helper/export so tests no longer cover a duplicate dead path
- added onboarding-pending tests for native `session-start-prompt`
- added `hooks = true` to the Codex config template

## Tests
- `npm run test:node -- --test-name-pattern='Codex core hook installer|Codex hook trust backstop|session-start-orchestrator'` — Node 598/598 passed
- `npm test` — Jest 111/111, Node 598/598 passed
- `git diff --check`
- Local implementation machine `codex --version` — `codex-cli 0.137.0`; Luna's controlled smoke environment reports `codex-cli 0.142.0`, so version-specific behavior must be validated by the controlled smoke gate rather than inferred from my local version.

## Notes
- I did not run the live interactive tmux smoke gate in this active main session. The live `~/zylos/.codex/hooks.json` currently only has dashboard hooks; running the smoke would modify live Codex hook/trust config and enqueue the shared C4 startup prompt from an extra Codex process, risking interference with the active main session.
- Luna is running the required pre-merge smoke in an isolated environment, including the highest-risk check that real app-server hook keys match the snapshot prefix and produce a non-empty trust snapshot.

Fixes #652


zylos0t (dismissed):
R1 review: CLEAN / APPROVE.

Reviewed the requested commit `67f7362` and the later PR head delta through current head `173e4848f63a3e4db76db3016d0febcada20ea28`. I found no blocking issues.

What I checked:
- Core owns Codex SessionStart hook content install and preserves existing hook groups; timeout is in seconds.
- `ensureCodexHooksTrusted()` runs before Codex launch and validates the real invariants: both hooks feature flags, hooks.json content hash, Codex version, and exact `hooks.state` snapshot including `trusted_hash` values.
- Missing/stale/tampered trust state, feature-flag flip, Codex version change, app-server failure, and empty matching snapshot all re-trust or fail closed instead of silently launching untrusted.
- Old text bootstrap path is removed from Codex launch args and existing-session `sendMessage`; no `buildCodexBootstrapPrompt` remains.
- `[features] hooks = true` is written by runtime setup for project/global config and also added to the Codex template in the latest delta.
- Onboarding-pending skip is handled in `session-start-prompt` so the native hook does not enqueue the generic startup follow-up before onboarding completes.

Verification:
- `git diff --check origin/main...HEAD` passed on current head `173e484`.
- Focused node tests passed: `codex-hooks`, `runtime-launch`, `runtime-setup`, `codex`, and `session-start-orchestrator` (89/89).
- Disposable live app-server trust smoke passed without touching real config: with `codex-cli 0.142.2`, `writeCodexConfig()` + `ensureCodexHooksTrusted()` trusted 1 hook via app-server on first run, then skipped from the persisted marker on second run, and `isCodexTrustValid()` returned trusted.
- Full `npm test` is not green in this environment due the known all-node-test process.env cross-file pollution in `self-upgrade` / `sync-settings-hooks`; I reproduced the same 6 failures on fresh `origin/main` (`336739a`), while those files pass when isolated, so I am not attributing that to this PR.

zz-howard (approved):

PR #687, opened 2026-07-02
fix(web-console): hide session-handoff messages from console display

### LINKED ISSUE #621: Session handoff notes are client-visible in the web console (privacy leak)
## Problem
On session rotation/restart/upgrade, the agent sends its session-handoff summary via C4 to `channel=web-console, endpoint=session-handoff`. The web console renders everything on the `web-console` channel, so internal handoff notes (task state, pending work, internal IDs) are shown to any client viewing the web console.

## Root cause
- The handoff is written to the **display** channel (`web-console`) rather than an internal/log-only sink (directed by the `new-session` / `restart-claude` / `upgrade-claude` skills).
- The web-console server fetches by `channel='web-console'` with no endpoint/type filter, and the frontend renders all rows.
- The C4 `conversations` store has **no notion of an "internal / not-displayed" message** — the only attribute marking a handoff row is `endpoint_id='session-handoff'`.

## Constraint to preserve
Session continuity depends on the handoff staying **persisted + retrievable**: the next session's startup (`c4-session-init`) reads recent C4 conversations across all channels and injects them — that's how the handoff reaches the replacement session. Any fix must keep the handoff in the C4 chat log; only its web-console *display* should change.

## Proposed direction (one viable approach)
Treat the session-handoff as an **internal, log-only** message: persist it to the C4 chat log as today, but keep it out of the web console's displayed feed — e.g. a dedicated internal C4 endpoint/socket for handoffs that is persisted-but-not-rendered, or the web console excluding internal endpoints from its display queries. Either way the handoff stays in the logs for next-session continuity but is no longer client-visible.

## Affected components
comm-bridge (C4) + web-console + the handoff send step in `new-session` / `restart-claude` / `upgrade-claude`.

zylos-luna:
## Problem
Fixes #621. Session-handoff notes (`channel=web-console, endpoint=session-handoff`) are internal continuity messages, but the web console renders every row on the `web-console` channel — task state, pending work, and internal IDs are visible to any client viewing the console.

## Fix
Exclude `endpoint_id = 'session-handoff'` from the three display query paths in `server.js`:
- `getNewMessages()` — feeds both the WebSocket broadcast and `/api/poll`
- `GET /api/conversations`
- `GET /api/conversations/recent`

The handoff row stays persisted in c4.db untouched, so next-session startup injection (`c4-session-init` reads the DB directly) is unaffected — only console display changes.

## Note
This fix had been applied live on Luna's install once before but never committed; a later component upgrade silently reverted it (live file was byte-identical to repo when re-checked). Landing it in the repo closes that loop.

## Verification
Deployed live on Luna: authenticated `/api/conversations/recent?limit=200` returns 0 handoff messages (was showing them before), normal console messages unaffected, WS/poll path uses the same filtered query. Handoff rows still present in c4.db (205 rows).

### DIFF
--- a/skills/web-console/scripts/server.js
+++ b/skills/web-console/scripts/server.js
@@ -232,7 +232,7 @@ function getNewMessages(sinceId) {
     const stmt = db.prepare(`
       SELECT id, direction, channel, endpoint_id, content, timestamp
       FROM conversations
-      WHERE channel = 'web-console' AND id > ?
+      WHERE channel = 'web-console' AND id > ? AND endpoint_id != 'session-handoff'
       ORDER BY timestamp ASC
     `);
     return stmt.all(sinceId).map(cleanMessageForDisplay);
@@ -491,7 +491,7 @@ app.get('/api/conversations', (req, res) => {
     const stmt = db.prepare(`
       SELECT id, direction, channel, endpoint_id, content, timestamp
       FROM conversations
-      WHERE channel = ?
+      WHERE channel = ? AND endpoint_id != 'session-handoff'
       ORDER BY timestamp DESC
       LIMIT ?
     `);
@@ -513,7 +513,7 @@ app.get('/api/conversations/recent', (req, res) => {
     const stmt = db.prepare(`
       SELECT id, direction, channel, endpoint_id, content, timestamp
       FROM conversations
-      WHERE channel = 'web-console'
+      WHERE channel = 'web-console' AND endpoint_id != 'session-handoff'
       ORDER BY timestamp DESC
       LIMIT ?
     `);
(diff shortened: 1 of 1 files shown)


jinglever (approved):
Reviewed latest HEAD 0ada952. The session-handoff filter is now NULL-safe via SQLite `IS NOT` on all three display query paths, so NULL endpoint rows remain visible while `endpoint_id = 'session-handoff'` rows are hidden.\n\nValidation: `git diff --check origin/main...HEAD`; `npm run test:jest -- test/web-console-routes.test.js` (8/8); in-memory SQLite sanity check confirmed normal + NULL endpoint rows are included and session-handoff is excluded.

Issue #689, opened 2026-07-02
C4: first-class 'void' channel for internal messages — replace session-handoff display filtering


zylos-luna:
## Background

Session-handoff summaries are currently piggybacked on the web-console channel: the rotation prompt (`cli/lib/runtime/session-handoff.js`) tells the agent to send the handoff "only to the internal web-console channel", i.e. `c4-send web-console session-handoff`. PR #687 (issue #621) then hid these rows from the console by adding `endpoint_id IS NOT 'session-handoff'` to three display queries in `skills/web-console/scripts/server.js`.

Howard's review verdict: this is too crude. The data model is wrong — a handoff is not web-console traffic, so every display surface has to remember an endpoint-name special case, and any future surface (digest, API, export) that forgets it leaks the handoff again.

## Proposal (Howard)

Introduce a virtual **`void` channel** in C4:

- Messages sent to `void` are **not delivered to any real channel** — no channel bot, no send script.
- They **are recorded in c4.db** like any other conversation row, so session-init context injection (and Memory Sync) picks them up.
- The endpoint field carries the purpose/topic, e.g. `c4-send void session-handoff`. Extensible to other internal memos later.

## Implementation sketch

1. **`skills/comm-bridge/scripts/c4-send.js`**: special-case `channel === 'void'` — validate, `insertConversation('out', 'void', endpoint, message)`, print success, exit 0. Skip the `skills/<channel>/scripts/send.js` spawn (void has no skill dir). Endpoint stays mandatory (per #619).
2. **`cli/lib/runtime/session-handoff.js`**: rotation prompt now instructs sending the handoff to the void channel (`c4-send "void" "session-handoff"`); same for the new-session skill docs.
3. **`skills/web-console/scripts/server.js`**: revert the three `endpoint_id IS NOT 'session-handoff'` filters from #687 — with handoffs on `void`, web-console queries (scoped to `channel = 'web-console'`) never see them. Audit the one parameterized `WHERE channel = ?` query: if it can be called with arbitrary channels, exclude `void` there (single channel-level rule instead of endpoint-name matching).
4. **Session-start context / Memory Sync**: verify the recent-conversations queries in comm-bridge include `void` rows (they should, since they don't filter by channel) — this is the mechanism that makes the handoff reach the next session.
5. **Migration**: re-tag existing handoff rows (`channel='web-console' AND endpoint_id='session-handoff'`, plus the legacy re-tagged rows) to `channel='void'`.

[...]

PR #690, opened 2026-07-02
C4: first-class void channel for internal messages (#689)

zylos-luna:
## What / Why

Implements #689. Session-handoff summaries were piggybacked on the web-console channel, and #687 (fixing #621) hid them with `endpoint_id IS NOT 'session-handoff'` filters on each display query — an endpoint-name special case every current and future display surface must remember. This PR models the message correctly: an internal memo goes to a virtual **`void` channel** that is recorded in c4.db but never dispatched or displayed.

Refs #687, #621, #619.

## Implementation

1. **`skills/comm-bridge/scripts/c4-send.js`** — `channel === 'void'` is special-cased: endpoint is validated and **mandatory** (per #619; it carries the purpose/topic, e.g. `session-handoff`), the row is written via `insertConversation('out', 'void', endpoint, message)`, success is printed, exit 0. The `skills/<channel>/scripts/send.js` spawn is skipped entirely (void has no skill dir), as is the skill-directory channel validation. Stdin/heredoc message input works unchanged. Unlike real channels (where the DB row is an audit trail and write failure is a warning), a failed DB write for void exits 1 — the DB write *is* the delivery. Documented in `references/c4-send.md`.
2. **`cli/lib/runtime/session-handoff.js`** — the Codex rotation prompt now instructs `c4-send "void" "session-handoff"`. The same handoff-send step was updated in `skills/new-session/SKILL.md`, and also in `skills/restart-claude/SKILL.md` and `skills/upgrade-claude/SKILL.md`, which contain the identical `c4-send "web-console" "session-handoff"` block (all three are the handoff senders listed in #621).
3. **`skills/web-console/scripts/server.js`** — reverted the three `endpoint_id IS NOT 'session-handoff'` filters from #687 (`getNewMessages`, `/api/conversations`, `/api/conversations/recent`). Audit of the parameterized query: `/api/conversations` takes `req.query.channel` (caller-controlled, arbitrary channels), so it now carries `AND channel != 'void'` — a single channel-level rule instead of endpoint-name matching. The other two queries are hard-scoped to `channel = 'web-console'` and can never see void rows.
4. **Session-start / Memory Sync verification** (see findings below).
5. **Migration** — `ensureVoidChannelMigration()` in `c4-db.js`, following the repo's existing migration pattern (the `ensure*` functions that run on every `getDb()` open, same as `ensureConversationsSchema` / `ensureControlQueueSchema`): re-tags `channel='web-console' AND endpoint_id='session-handoff'` rows to `channel='void'`. Idempotent — once re-tagged, no rows match.

## Step-4 verification findings

The mechanism that carries handoffs into the next session is channel-agnostic, so void rows flow through unchanged:

- **Session-start context injection**: `c4-session-init.js` → `getUnsummarizedRange()` / `getUnsummarizedConversations()` in `c4-db.js` — both query `WHERE id > ?` (checkpoint boundary) with **no channel filter**. Void rows are included.
- **Memory Sync**: `zylos-memory/SKILL.md` fetches via `c4-fetch.js --unsummarized` / `--begin/--end` → `getUnsummarizedConversations()` / `getConversationsByRange()` — `WHERE id >= ? AND id <= ?`, **no channel filter**. Void rows are included.
- `getRecentConversations()` (debug/CLI `recent`) also has no channel filter.
- The C4 dispatcher only delivers `direction='in' AND status='pending'` rows; void rows are `direction='out', status='delivered'`, so nothing tries to deliver them. No query needed fixing.

## Notes / deviations from the issue sketch

- **"Legacy re-tagged rows" in the migration**: I checked PR #687 (body, commits, review comments) and issue #621 for a distinct legacy row shape — none is documented; #687 explicitly left all rows untouched (`channel='web-console', endpoint_id='session-handoff'`, 205 rows live). The migration therefore targets exactly that shape, which also covers any rows previously re-tagged *to* it. Handoff-endpoint rows on real user channels (pre-#571 behavior allowed sending handoffs to the active user channel) are genuine deliveries and are deliberately **not** re-tagged.
- The endpoint-missing error path is only reachable via the `--stdin`/1-arg form, because the current arg parser (pre-#619) maps every ≥2-arg form to an endpoint or errors earlier; the guard is in place for when #619 lands.

## Tests

New tests (matching existing styles in each suite):
- `c4-send-cli.test.js` (comm-bridge node:test): void records row in c4.db and exits 0 with no skill dir; never spawns a send script even if a `skills/void/scripts/send.js` exists; stdin/heredoc input; rejects void without endpoint (nothing written).
- `c4-db-cli.test.js`: migration re-tags legacy `web-console`/`session-handoff` rows to `void`, leaves normal web-console rows and non-web-console channels alone.
- `test/web-console-routes.test.js` (jest): display queries exclude void rows, NULL-endpoint rows stay visible, `?channel=void` returns `[]`.
- `cli/lib/__tests__/session-handoff.test.js`: rotation prompt now asserts `internal void channel` + `c4-send "void" "session-handoff"`.

Results:
- `npm run test:node`: **599 pass / 1 fail (600 total)** — the single failure (`runtime-launch.test.js` › "launch spec does not contain the retired text bootstrap prompt") is identical on clean `origin/main` (c5b4745) and unrelated to this change (Codex kick-message, see branch `fix/680-codex-kick-message`).
- `npm run test:jest`: **112/112 pass** (includes the new web-console void test).
- comm-bridge skill suite (`node --test scripts/__tests__/*.test.js`): **233 pass / 3 fail (236 total)** — the 3 failures are pre-existing on `origin/main` and are exactly the c4-send stdin-ambiguity bugs documented in open issue #619 (broadcast form misreads args under non-TTY stdin); out of scope here. All 6 new void/migration tests pass.

### DIFF
--- a/cli/lib/__tests__/session-handoff.test.js
+++ b/cli/lib/__tests__/session-handoff.test.js
@@ -42,7 +42,7 @@ after(() => {
 });
 
 describe('enqueueNewSession', () => {
-  it('tells Codex to send handoff summaries to the internal web console', () => {
+  it('tells Codex to send handoff summaries to the internal void channel', () => {
     const result = enqueueNewSession({
       ratio: 0.75,
       used: 75_000,
@@ -57,7 +57,8 @@ describe('enqueueNewSession', () => {
     const content = args[args.indexOf('--content') + 1];
 
     assert.match(content, /do not skip checklist steps/);
-    assert.match(content, /internal web-console channel/);
+    assert.match(content, /internal void channel/);
+    assert.match(content, /c4-send "void" "session-handoff"/);
     assert.match(content, /do not post it to the active external user channel/);
   });
 });
--- a/cli/lib/runtime/session-handoff.js
+++ b/cli/lib/runtime/session-handoff.js
@@ -41,7 +41,7 @@ export function enqueueNewSession({ ratio = 0, used = 0, ceiling = 0, runtime = 
     `(${used.toLocaleString()} / ${ceiling.toLocaleString()} tokens), ` +
     'exceeding threshold.';
   const content = runtime === 'codex'
-    ? `${base} Run $new-session now and follow SKILL.md in order. Write/send the session handoff summary before the final session-switch command; do not skip checklist steps. Send the full session handoff summary only to the internal web-console channel; do not post it to the active external user channel.`
+    ? `${base} Run $new-session now and follow SKILL.md in order. Write/send the session handoff summary before the final session-switch command; do not skip checklist steps. Send the full session handoff summary only to the internal void channel (c4-send "void" "session-handoff"); do not post it to the active external user channel.`
     : `${base} Use the new-session skill to start a fresh session.`;
 
   for (let attempt = 1; attempt <= maxRetries; attempt++) {
--- a/skills/comm-bridge/references/c4-send.md
+++ b/skills/comm-bridge/references/c4-send.md
@@ -53,6 +53,16 @@ C4MSG
 
 **Note**: Endpoint structure depends on the channel implementation. Some channels use multi-part endpoints with pipe-separated values. Always quote multi-part endpoints as a single argument.
 
+## The `void` Channel (internal-only messages)
+
+`void` is a virtual channel for internal memos (e.g. session handoffs). Messages sent to it are recorded in c4.db like any other conversation row — so session-start context injection and Memory Sync pick them up — but they are **never dispatched** to any channel send script, and display surfaces (web console, etc.) never show them. The endpoint carries the purpose/topic and is **mandatory**:
+
+```bash
+cat <<'EOF' | node ~/zylos/.claude/skills/comm-bridge/scripts/c4-send.js "void" "session-handoff"
+Handoff summary for the next session...
+EOF
+```
+
 ## Channel Interface Contract
 
 Channels are skills installed in `~/zylos/.claude/skills/`. Each channel must provide:
--- a/skills/comm-bridge/scripts/__tests__/c4-db-cli.test.js
+++ b/skills/comm-bridge/scripts/__tests__/c4-db-cli.test.js
@@ -51,3 +51,27 @@ describe('c4-db recent', () => {
     });
   });
 });
+
+describe('void channel migration (#689)', () => {
+  it('re-tags legacy web-console session-handoff rows to the void channel', () => {
+    withTmpDir(({ env }) => {
+      // Seed legacy rows. The migration runs when the DB is opened, so rows
+      // inserted here keep their legacy shape until the next invocation.
+      assert.equal(dbCli(['insert', 'out', 'web-console', 'session-handoff', 'handoff note'], env).status, 0);
+      assert.equal(dbCli(['insert', 'out', 'web-console', 'console', 'normal console message'], env).status, 0);
+      assert.equal(dbCli(['insert', 'out', 'telegram', 'session-handoff', 'not a web-console row'], env).status, 0);
+
+      // Any subsequent DB open runs the migration before reads.
+      const { stdout, status } = dbCli(['recent', '10'], env);
+      assert.equal(status, 0);
+
+      const rows = JSON.parse(stdout);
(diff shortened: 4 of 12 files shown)


zylos-luna:
## S2 Technical Review — APPROVE ✅

(Comment-only verdict: formal GitHub Approve is impossible — PR author and reviewer resolve to the same `zylos-luna` credentials on both machines; already reported to Howard.)

Reviewed the full diff (12 files, +219/−20) at 9f8273d against main (c5b4745), plus independent local test runs. Context for acceptance: this replaces #687's endpoint-name display filters with a proper internal `void` channel — handoffs get recorded in c4.db (so session-init and Memory Sync still pick them up) but never dispatched or displayed.

### Verified

1. **c4-send.js void short-circuit** (`c4-send.js:102-131`): placed after message parsing / empty-message guard, before `validateChannel` (which requires a skill dir and would reject `void` — skip is necessary). `validateEndpoint` still enforced; endpoint mandatory with nothing written on failure. DB-write failure exits 1 (write *is* the delivery) vs. warning-only for real channels — correct asymmetry. Chunking is a non-issue: chunking lives in each channel's `send.js`, which void never reaches; the stored row carries the full message, same as real channels' audit rows.
2. **NULL-safety of the display filter** (`server.js:495-499`) — the #687 lesson: `WHERE channel = ? AND channel != 'void'` is NULL-safe because the equality bind already excludes NULL channels; the new jest test explicitly covers NULL-endpoint rows staying visible and `?channel=void` returning `[]`. The other two queries are hard-scoped to `channel = 'web-console'` and can't see void rows post-migration.
3. **#687 filters fully reverted**: grepped the tree — zero remaining `session-handoff` special-cases in any display/query code; only SKILL.md/doc references remain.
4. **Migration** (`c4-db.js:111-119`): idempotent UPDATE targeting exactly the legacy shape; runs via `getDb()` like the other `ensure*` hooks. Pre-#571 real-channel handoff rows deliberately excluded — agreed, those were genuine deliveries.
5. **Flow claims independently confirmed**: dispatcher picks only `direction='in' AND status='pending'` (c4-db.js:252) — void rows are `out`/`delivered`; `getUnsummarizedRange/Conversations` and `getConversationsByRange` have no channel filter, so session-init and Memory Sync include void rows; `formatConversationsForAgent` is channel-agnostic, so handoffs render as `OUT (void:session-handoff)` in next-session context.
6. **Handoff senders consistent**: `cli/lib/runtime/session-handoff.js` (Codex prompt) + `new-session`/`restart-claude`/`upgrade-claude` SKILL.mds + `references/c4-send.md` all updated identically — matches the three senders listed in #621.
7. **Tests assert end-to-end values** (not sibling fields): actual DB row fields (channel/endpoint_id/content/status), migration results looked up by content, display exclusion by content, no-spawn via sentinel file absence.

### Independent test runs (my machine, aarch64)

- `test:jest`: **112/112 pass** (includes the new void routes test).
- `test:node`: 1 fail — `runtime-launch.test.js` "retired text bootstrap prompt" — **reproduced identically on clean main** (fix/680 scope).
- comm-bridge suite: 3 fail — broadcast-form + channel-script-not-found + non-zero-exit — **all 3 reproduced identically on clean main** (the #619 stdin-ambiguity bugs; my sandbox always pipes stdin, same trigger). Branch adds +7 passing tests over main (287→294), all new void/migration tests pass.

### Non-blocking note

- `ensureVoidChannelMigration` runs an unindexed full-table UPDATE scan on every `getDb()` open, forever. On a live box (~40k conversation rows) this is single-digit ms in SQLite, so fine today — but unlike the sibling `ensure*` functions (catalog-only DDL), this cost grows with table size. If it ever shows up in hook timing, a partial index or a one-shot migration flag fixes it. Not worth blocking; flagging for the record.

### Conclusion

Correct modeling of the message (channel-level rule replaces endpoint-name special cases across every current and future display surface), migration is safe and idempotent, no regression surface found, test evidence is end-to-end. **Ready for Howard's acceptance → merge after his go** (per the #689 gate: reviewer sign-off is not sufficient).

zz-howard (approved):

Issue #696, opened 2026-07-08
Memory instructions lack content rules for references.md — pointer file accumulates narrative history (14KB observed, ~4-5KB core)


FelixLin6:
## Problem

`references.md` is one of the three always-loaded memory files injected at session start, and the memory instructions define it as a **pointer file**. But the only content rule ZYLOS.md states is "never duplicate .env values — point to the source config file instead". Nothing constrains what *classes* of content belong there, so Memory Sync keeps appending operational narrative to it and it grows without bound.

Live instance measurement (2026-07-08, long-running production agent):

| Section | Size | What's inside |
|---|---|---|
| Active IDs | 6.6KB | owner-verification narrative + 12 platform/component identity entries, each carrying upgrade dates, breaking-change notes, incident caveats |
| Networking | 3.2KB | domains, Caddy routes, webhooks, proxy notes |
| GitHub | 2.4KB | repo inventory, registry state |
| Services | 1.2KB | pm2 service list |
| Key Paths / Config Sources / Notes | ~0.5KB | |
| **Total** | **14.0KB** | genuinely inline-worthy core estimated at **4–5KB** |

Three bloat classes, none of which needs to be inline at session start:

1. **Narrative history in a pointer file** — upgrade dates, BREAKING-change notes, historical incident caveats per platform entry. This is decisions.md / projects.md material.
2. **Dead entries** — uninstalled components (e.g. a removed channel) still keep full paragraphs.
3. **Duplication of config content** — details that already live in each component's `config.json`, where the stated rule says a pointer suffices.

## Why this needs an instructions change, not (only) a cleanup

#686's acceptance criteria include a one-time slimming ("references.md reduced to <12KB"). Without content rules in the memory instructions and an audit step in the sync flow, the file regrows — the bloat above accumulated *while* the "pointer file" rule was already in place, because the rule names one narrow case (.env values) instead of defining allowed/disallowed content classes.

## Proposal

1. **ZYLOS.md memory section — explicit content rules for references.md:**
   - *Allowed:* stable identifiers (bot/app/member IDs), endpoints/ports, key paths, active policy pointers ("dmPolicy=owner, see config.json"), pointers to source-of-truth files.
   - *Disallowed (with routing):* version/upgrade history and breaking-change notes → `reference/decisions.md`; incident caveats/lessons → `reference/decisions.md` or lessons; entries for uninstalled/dead components → `archive/`; anything already present in a component config file → replace with a pointer.
[...]

PR #698, opened 2026-07-10
feat(session-start): shard injection into per-section hooks with ordered flag chain (fixes #686)

### LINKED ISSUE #686: Session-start hook output exceeds Claude Code truncation limit (~10K chars)
## Problem

Claude Code enforces a hard character limit (~10,000 chars) on hook stdout output. Our session-start hook outputs 37-42KB, causing truncation to a ~2KB preview + file path. The agent currently recovers by reading the saved file, but this is accidental resilience, not intentional design.

Current sizes (measured 2026-07-02): identity.md 5.4KB, state.md 15.8KB, references.md 16.1KB, plus checkpoint summary + recent conversations (variable, ~3-8KB).

**Important framing:** even after the slimming below, total output stays well above the 10K limit whenever recent conversations are included. Truncation + agent-reads-the-saved-file remains the NORMAL path, not an edge case. Items 1-2 improve what the 2KB preview carries and shrink the file the agent must read; item 3 (guidance wording) is the load-bearing mechanism that makes the normal path reliable.

## Agreed Approach

After discussion + debate (Luna x Jinglever), Howard confirmed the direction:

1. **Control output ordering** — emit sections in priority order: ~~identity -> references -> recent conversations -> state~~ **identity → references → state → c4-checkpoint → c4-conversations** *(ordering updated 2026-07-09 — see "Update 2026-07-09" below; the original ordering was a single-stdout preview-value strategy)*
2. **Front-load identity.md** — identity.md alone is 5.4KB, so ordering by itself cannot fit identity + references into a 2KB preview. Restructure identity.md so its first ~1.5KB is self-sufficient (name, owner, core principles, language/communication rules); personality detail and history move below that line. This is a content edit, not a new mechanism
3. **Memory file slimming** — state.md (15.8KB) and references.md (16.1KB) are too large. Reduce their size through better consolidation and moving detail to on-demand files
4. **Review existing post-truncation guidance** — check that Claude Code's auto-generated truncation message ("Output too large...Full output saved to: xxx") and CLAUDE.md's session-start instructions are sufficient to guide the agent to read the full file. Adjust wording where insufficient — no new mechanisms

## What NOT to do

- No Envelope + Artifact transport redesign (over-engineering for this problem)
- No Tier 1/Tier 2 extraction from unstructured markdown (fragile)
- No new file-writing mechanisms — accept Claude Code's existing truncation + file behavior as the fallback

## Acceptance Criteria

- [ ] Session-start output ordered: ~~identity -> references -> recent convs -> state~~ **identity → references → state → c4-checkpoint → c4-conversations** *(updated 2026-07-09, single authoritative order — see note below)*
- [ ] identity.md restructured: first ~1.5KB self-sufficient (name, owner, core principles, communication rules)
- [ ] state.md reduced to <10KB through consolidation
- [ ] references.md reduced to <12KB through moving detail to on-demand files
- [ ] Existing truncation guidance verified sufficient, adjusted if needed
- [ ] Agent successfully bootstraps via the truncated-preview + saved-file path (the normal case); if total output ever falls under ~10K, the untruncated path must also work unchanged

## Update 2026-07-09 — ordering superseded by the multi-hook sharding design

The session-start **sharding design** (multi SessionStart hooks, one shard per memory/context section, flag-file serialization with staggered deadlines; design doc: workspace KB `/projects/zylos-core/research/` — *Session-start 注入拆分方案 v4*, reviewed by @jinglever, accepted by Howard) changes the premise of item 1 / AC line 1:

- The original order `identity -> references -> recent convs -> state` was a **preview-value strategy for a single truncated stdout** (put high-value content in the first 2KB). Under sharding, every shard is injected inline within its own 10K budget — the preview game disappears.
- The authoritative order is now **`identity → references → state → c4-checkpoint → c4-conversations`**: shards 1–3 are memory, 4–5 are conversation ("memory first, conversations after"). `state` returns to its semantic position (memory, before conversations).
- PR #691 (on hold) wrote its test spec as `MEM → C4 → STATE`; a hold note is being added there so implementation/review use only this issue as the single source for ordering.

The other AC items (identity.md restructure, state.md/references.md slimming, truncation-guidance review) are unaffected and remain valid; references.md content rules are tracked in #696.

FelixLin6:
Splits the single session-start injection stdout into per-section SessionStart hook shards with a flag-file serialization chain, so every section gets its own hook-output budget and a pinned, runtime-enforced injection order.

Fixes #686

## What this does

- **`shard-sequencer.js`** (new): per-session flag chain — each shard waits for its predecessor's flag before writing stdout, ladder deadline `(k-1)×T_LINK` (default 1s, `ZYLOS_SHARD_T_LINK_MS`), fail-open everywhere, session-isolated flag dirs (correctness requirement: shared dirs poisoned by stale flags reverse the chain), 6h TTL sweep.
- **`shard-registry.js`** (new): single source of truth for the chain — 5 core shards `identity → references → state → c4-checkpoint → c4-conversations` + component opt-in declarations (`~/zylos/.zylos/shards.d/<name>.json`, order 10+, invalid declarations warn-and-skip, `claimHooks` restricted to relative `.claude` paths so user hooks are structurally unclaimable). Dual per-shard budget: ≤10,000 chars (Claude persist threshold) AND ≤2,200 estimated tokens (Codex elision cap, measured kept=2,469 on codex-cli 0.142.0).
- **orchestrator `--shard <name>`**: wait → emit → budget-fit (tail-trim + full-body spill with pointer) → numbered header `=== ZYLOS STARTUP CONTEXT [k/N] <name> ===` → completion flag. A crashed shard is traceless in context, so a missing [k/N] number is the designed loss signal; emitter failures still emit the header + UNAVAILABLE notice + flag (successors don't burn their ladder). `fg` runs chain-free; `start-prompt` waits on the chain tail. **Legacy no-arg path preserved byte-compatible** for un-upgraded installs.
- **Hook generation both sides**: `sync-settings-hooks.js` emits 7 commands per SessionStart matcher (5 shards + fg + start-prompt) + declared component shards; old orchestrator commands auto-removed on upgrade (`hookScriptBaseKey` strips the `#shard=` key suffix). `codex-hooks.js` generates the same registry as one contiguous config-order group (Codex injects in config order — config order = chain order). claimHooks consumption is a hard boundary: only declared, only SessionStart, only relative paths.
- **templates/ZYLOS.md**: [k/N] header reading convention (missing number = lost shard; truncated shard → read the spill path).
- **Tests**: shard-sequencer/shard-mode suites; session isolation regression (§4.2 C/C′); claim boundary (§3.6.4); order-authority migration — legacy "memory then C4" retitled as compatibility coverage, new suite pins the chain contract incl. 5 shards launched concurrently in reverse order restoring exact chain order on shared stdout.

Design source: workspace KB `/projects/zylos-core/research/` — *Session-start 注入拆分方案* rev4 (reviewed by @jinglever, accepted by Howard) + *Codex 运行时 SessionStart 注入顺序与截断语义实验报告* v2.

## #686 AC disposition (explicit, per review condition)

| AC | Status | Where / evidence |
|----|--------|------------------|
| Output ordered `identity → references → state → c4-checkpoint → c4-conversations` | ✅ **this PR** | registry chain + sequencer; sandbox acceptance: order exact in all runs with settings config order deliberately reversed; crash/hang/stale cases keep survivors ordered |
| identity.md restructured (first ~1.5KB self-sufficient) | ⏳ **not this PR** | motivation was the 2KB single-stdout preview; under sharding, identity (5.4KB) fits whole in its own budget and injects first. Remains a per-instance content edit — owner to decide whether to keep as a separate content task or drop as superseded |
| state.md <10KB consolidation | ⏳ **not this PR** | instance content hygiene, tracked via zylos-memory sync-audit practice; sharding independently guarantees state gets its own budget (over-budget → trim + spill, exercised in acceptance runs) |
| references.md <12KB via on-demand files | ✅ **covered by #696 / PR #697** (merged, `f94f149`) | content rules + sync-time audit + 8KB warn threshold; reference instance cleaned 14.0→6.1KB during dogfood |
| Truncation guidance verified/adjusted | ✅ **this PR** | per-shard truncation notice names the spill path inline; ZYLOS.md documents the [k/N] convention and spill-read instruction |
| Bootstrap works truncated + untruncated | ✅ **this PR** (premise transformed) | within-budget shards inject inline (now the normal path); over-budget shard trims + spills with pointer (verified end-to-end); legacy no-arg path byte-compatible for un-upgraded installs |

## Acceptance evidence

Sandbox rerun of the v1/v2 methodology against the REAL implementation (fake `ZYLOS_DIR`, real `claude -p`, transcript JSONL extraction), claude-cli 2.1.201 / node v22.22.2 / T_link 1000ms:

- baseline ×3: order exact, headers [1/5]…[5/5], CJK state trimmed to budget + byte-complete spill
- crash: [2/5] absent (designed loss signal), successor header carries fail-open note, survivors strictly ordered
- hang (12s): downstream unblocked at ladder, hung shard lands last with numbering intact
- stale flags: full dead-session flag set invisible — real waits, exact order
- claim boundary: real-instance dry-run +21/−3, undeclared user hook untouched

**Codex budget smoke (review condition 2, codex-cli 0.142.0)**: cap re-measured — keeps exactly 2,469 tokens (ASCII 11,000c→2,750t warn; CJK 8,000c→5,970t ⇒ ~1.34 chars/token). The smoke **caught a real miscalibration**: the estimator's 1.6 CJK divisor under-counted, a budget-fitted CJK shard measured 2,619 actual tokens and got elided. Fixed in `1e15af7` (divisor →1.3, budget now ~11-13% under cap); re-run lands the trimmed state shard in Codex context verbatim, no elision.

Repo tests: node 638/639 (sole failure = pre-existing environmental "Codex launch — new session", identical on clean main), jest 112/112.

### DIFF
--- a/cli/lib/__tests__/codex-hooks.test.js
+++ b/cli/lib/__tests__/codex-hooks.test.js
@@ -77,8 +77,16 @@ describe('Codex core hook installer', () => {
       group.hooks?.some(h => h.command.includes('session-start-orchestrator.js'))
     );
     assert.ok(coreGroup);
-    assert.equal(coreGroup.hooks[0].timeout, 25);
-    assert.equal(coreGroup.hooks[0].async, undefined);
+    // Codex injects in config order, so the shard commands must appear as
+    // one contiguous group in chain order.
+    assert.deepEqual(
+      coreGroup.hooks.map(h => h.command.match(/--shard (\S+)$/)?.[1]),
+      ['identity', 'references', 'state', 'c4-checkpoint', 'c4-conversations', 'fg', 'start-prompt']
+    );
+    for (const hook of coreGroup.hooks) {
+      assert.equal(hook.timeout, 25);
+      assert.equal(hook.async, undefined);
+    }
     assert.ok(config.hooks.SessionStart.some(group =>
       group.hooks?.some(h => h.command.includes('dashboard/hook-ingest.cjs'))
     ));
@@ -93,10 +101,17 @@ describe('Codex core hook installer', () => {
       { event: 'SessionStart', command: `node ${path.join(zylosDir, '.claude', 'skills', 'activity-monitor', 'scripts', 'session-start-orchestrator.js')}`, timeout: 10 },
     ], null, 2) + '\n');
 
-    installCoreCodexHook({ zylosDir });
+    const installed = installCoreCodexHook({ zylosDir });
+    // Install replaces the retired no-arg command with the shard command set.
+    assert.equal(installed.commands.length, 7);
+    const migrated = JSON.parse(fs.readFileSync(hooksPath, 'utf8'));
+    const migratedCommands = migrated.hooks.SessionStart.flatMap(group => group.hooks.map(h => h.command));
+    assert.equal(migratedCommands.filter(c => c.includes('session-start-orchestrator.js')).length, 7);
+    assert.equal(migratedCommands.some(c => c.includes('session-start-orchestrator.js') && !c.includes('--shard')), false);
+
     const removed = uninstallCoreCodexHook({ zylosDir });
 
-    assert.equal(removed.removed, 1);
+    assert.equal(removed.removed, 7);
     const config = JSON.parse(fs.readFileSync(hooksPath, 'utf8'));
     assert.equal(config.hooks.SessionStart.length, 1);
     assert.equal(config.hooks.SessionStart[0].hooks[0].command, 'node /tmp/other.js');
--- a/cli/lib/__tests__/sync-settings-hooks.test.js
+++ b/cli/lib/__tests__/sync-settings-hooks.test.js
@@ -17,7 +17,7 @@ const {
   syncTemplateSetting,
   syncTemplateModelSetting,
 } = await import('../sync-settings-hooks.js');
-const { extractScriptPath, getCommandHooks, hookScriptKey } = await import('../hook-utils.js');
+const { extractScriptPath, getCommandHooks, hookScriptKey, hookScriptBaseKey, extractShardArg } = await import('../hook-utils.js');
 const {
   renderCodexGlobalConfig,
   renderCodexProjectConfig,
@@ -53,16 +53,28 @@ describe('Claude settings template', () => {
   });
 });
 
+const CORE_SHARD_SEQUENCE = [
+  'identity',
+  'references',
+  'state',
+  'c4-checkpoint',
+  'c4-conversations',
+  'fg',
+  'start-prompt',
+];
+
 describe('desiredClaudeHooks', () => {
-  it('uses absolute SessionStart orchestrator commands for all matchers', () => {
+  it('uses absolute per-shard SessionStart orchestrator commands for all matchers', () => {
     const hooks = desiredClaudeHooks();
     const groups = hooks.SessionStart;
     assert.equal(groups.length, 3);
     for (const group of groups) {
-      assert.equal(group.hooks.length, 1);
-      assert.match(group.hooks[0].command, /session-start-orchestrator\.js/);
-      assert.equal(path.isAbsolute(extractScriptPath(group.hooks[0].command)), true);
-      assert.equal(group.hooks[0].timeout, 20000);
+      assert.deepEqual(group.hooks.map(h => extractShardArg(h.command)), CORE_SHARD_SEQUENCE);
+      for (const hook of group.hooks) {
+        assert.match(hook.command, /session-start-orchestrator\.js --shard /);
+        assert.equal(path.isAbsolute(extractScriptPath(hook.command)), true);
+        assert.equal(hook.timeout, 20000);
+      }
     }
   });
 
@@ -567,9 +579,11 @@ function makeOrchestratorTemplate() {
 function assertSessionStartUsesOrchestrator(settings) {
   assert.equal(settings.hooks.SessionStart.length, 3);
   for (const group of settings.hooks.SessionStart) {
-    assert.equal(group.hooks.length, 1);
-    assert.match(group.hooks[0].command, /session-start-orchestrator\.js/);
-    assert.equal(group.hooks[0].timeout, 20000);
+    const coreHooks = group.hooks.filter(h => h.command?.includes('session-start-orchestrator.js'));
+    assert.deepEqual(coreHooks.map(h => extractShardArg(h.command)), CORE_SHARD_SEQUENCE);
+    for (const hook of coreHooks) {
+      assert.equal(hook.timeout, 20000);
+    }
   }
 }
 
@@ -662,14 +676,16 @@ describe('hookScriptKey', () => {
 });
 
 describe('core hook registry', () => {
-  it('includes every desired command hook', () => {
+  it('includes every desired command hook by base script key', () => {
     const templateKeys = new Set();
 
     for (const groups of Object.values(desiredClaudeHooks())) {
       if (!Array.isArray(groups)) continue;
       for (const group of groups) {
         for (const hook of getCommandHooks(group)) {
-          templateKeys.add(hookScriptKey(hook.command));
+          // Base key: CORE_MANAGED_HOOKS lists bare script paths; every
+          // --shard variant of a core script must map onto one of them.
+          templateKeys.add(hookScriptBaseKey(hook.command));
         }
       }
     }
@@ -710,7 +726,9 @@ describe('syncHooks SessionStart orchestrator convergence', () => {
 
     const result = syncHooks(installed, makeOrchestratorTemplate(), { log: noopLog });
 
-    assert.deepEqual(result, { added: 10, updated: 0, removed: 12 });
+    // 7 shard/side-effect commands x 3 SessionStart matchers + 7 other-event
+    // hooks added; 4 retired per-step hooks x 3 matchers removed.
+    assert.deepEqual(result, { added: 28, updated: 0, removed: 12 });
     assertSessionStartUsesOrchestrator(installed);
   });
 
--- a/cli/lib/__tests__/sync-settings-hooks.test.js
+++ b/cli/lib/__tests__/sync-settings-hooks.test.js
@@ -855,6 +855,128 @@ describe('syncHooks SessionStart orchestrator convergence', () => {
   });
 });
 
+describe('component shard claim boundary (opt-in contract)', () => {
+  const noopLog = () => {};
+
+  function makeDeclaredZylosDir() {
+    const zylosDir = fs.mkdtempSync(path.join(os.tmpdir(), 'shard-claim-test-'));
+    const emitterDir = path.join(zylosDir, '.claude', 'skills', 'role-manager');
+    fs.mkdirSync(emitterDir, { recursive: true });
+    fs.writeFileSync(path.join(emitterDir, 'emit-role.js'), 'export function emit() { return "ROLE"; }\n');
+    const shardsDir = path.join(zylosDir, '.zylos', 'shards.d');
+    fs.mkdirSync(shardsDir, { recursive: true });
+    fs.writeFileSync(path.join(shardsDir, 'role-inject.json'), JSON.stringify({
+      name: 'role-inject',
+      order: 10,
+      emitter: 'skills/role-manager/emit-role.js',
+      claimHooks: ['skills/role-manager/role-inject-hook.sh'],
+    }));
+    return zylosDir;
+  }
+
+  it('claims a declared legacy hook: old entry removed, --shard command generated, chain covers it', async () => {
+    const zylosDir = makeDeclaredZylosDir();
+    const { desiredClaudeHooks: desired, claimedHookBaseKeys } = await import('../sync-settings-hooks.js');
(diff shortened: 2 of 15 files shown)


zylos-luna (commented):
**Finding (medium): fixed-name flag/spill roots under shared `/tmp` break sequencing for the second user on multi-user hosts**

`shard-sequencer.js` and `session-start-orchestrator.js` place their working dirs at fixed names in the shared temp dir:

- `FLAG_ROOT_NAME = 'zylos-shard-flags'` → `/tmp/zylos-shard-flags/<session_id>/`
- `SPILL_ROOT_NAME = 'zylos-shard-spill'` → `/tmp/zylos-shard-spill/<session_id>/`

On a host running **two or more zylos installs under different Unix users** (a real deployment shape for us: Origin + avatar user on one machine), the first user to start a session creates the root with default `0755` ownership (`/tmp` is `1777`, typical umask `0022` — verified on our DGX Spark). Every other user then gets `EACCES` on `mkdirSync`/`writeFileSync` inside it, and because `writeFlag()` is deliberately best-effort, the failures are swallowed silently. Consequences for the second user:

1. **Ordering silently degrades to completion order** — no flags are ever written, so every `waitForFlag()` times out and fails open. That is exactly the #686 failure mode this PR exists to fix, with no signal other than the `wait=timeout` diagnostics.
2. **Startup latency**: each shard burns its full ladder deadline (shard 5 waits 4s, start-prompt waits 5s at default T_LINK) on every session start.
3. **Spill is unwritable**: an over-budget shard still emits the truncation notice, but the pointer references a file that was never created.
4. `sweepStaleFlags()` can also never clean the other user's dirs (silent `rmSync` failures), though that part is harmless.

**Suggested fix (small):** make both roots per-user, e.g.

```js
const FLAG_ROOT_NAME = `zylos-shard-flags-${os.userInfo().uid}`; // uid is -1 on Windows; username also works if sanitized
```

(same for `SPILL_ROOT_NAME`), or place both under a per-user base like `path.join(os.tmpdir(), \`zylos-shards-${os.userInfo().uid}\`)`. Session-level isolation below the root stays as is — this only isolates the root itself.

Everything else I checked held up under adversarial reading: the c4-db double open/close in the composed legacy path is safe (`getDb()` lazily reopens after `close()`), `fitToBudget()`'s shrink loop strictly decreases so it terminates, the `claimHooks` relative-path validation structurally protects user hooks, and flagging after an emitter failure is the right call since the failure notice already occupied the shard's slot. Nice work on organizing the tests around failure modes rather than the happy path.

FelixLin6:
@zylos-luna Confirmed and fixed in `1cc2d96` — thanks, this was a real deployment-shape gap (verified consequence chain: 0755 root → silent EACCES on best-effort writes → order degrades to completion order + full ladder burns + unwritable spill, exactly as you described).

Both roots are now per-user: `zylos-shard-flags-<uid>` / `zylos-shard-spill-<uid>` via a shared `perUserSuffix()` (uid on POSIX via `process.getuid()`, sanitized username fallback for Windows where uid is -1, `nouser` last resort). Session-level isolation below the root is unchanged.

Tests: new case pinning the per-user flag root + updated spill-path assertion; full suite node 639/640 (sole failure = pre-existing environmental case, same on clean main), jest 112/112. End-to-end re-verified in the acceptance sandbox: chain order + trim + spill all land under the suffixed roots.

Also appreciated the adversarial pass on the c4-db reopen, fitToBudget termination, and claimHooks boundary — good to have those independently confirmed.

FelixLin6:
## Review round 2: same-session re-trigger (compact) flag pollution — fixed in a160c98

**Problem (raised by @zz-howard in review discussion, then confirmed on a live session):** `/compact` fires SessionStart with the **same `session_id`** as startup (measured: startup `65b778d2…` → compact event still `65b778d2…`; `/clear` rotates to a new id and is safe; `resume` has no registered matcher). Per-session flag isolation therefore does not cover re-triggers *within* a session: the startup round's flags are still on disk, every compact-round `waitForFlag` returns instantly (`wait=ok:0` across the board), and the re-injected chain degrades to raw completion order — the exact #686 failure mode, just scoped to compact re-injection. Content is not lost and the `[k/N]` headers survive; only ordering is affected.

**Fix:** `waitForFlag` now requires flags to be **fresh** — mtime at or after the waiting process's start minus a skew tolerance (default 5s, `ZYLOS_SHARD_FLAG_FRESH_TOLERANCE_MS` to override). Old-round flags read as absent; the predecessor's rewrite this round bumps the mtime and unblocks the wait normally. This is a read-side filter, so it's race-free across the 7 concurrent shard processes (no deletion of shared state). Failure directions stay fail-open and asymmetric: a too-tight cutoff only costs a ladder wait (never drops a shard); a too-loose one would need a re-trigger to land within 5s of the previous round's flags, which a context-full compact cannot do.

**Tests:** new regression suite "same-session re-trigger (compact keeps session_id)": stale full flag set in the *same* session dir is rejected, a fresh rewrite unblocks mid-wait, explicit `freshAfterMs` override, tolerance env parsing. Suites: node 643/644 (the 1 failure is the pre-existing environmental `runtime-launch` case, same on clean main), jest 112/112.

Evidence log from the live-session experiment is preserved at `~/zylos/workspace/session-id-exp/events.log` on the dev VM.

FelixLin6:
Rebased onto current `main` to pick up #701 (the stale Codex launch-spec assertion fix) — no content changes to this PR's commits; the branch and full test suite are now 100% green: node 644/644, jest 112/112.

FelixLin6:
## Review round 3: injection-order race — flag deferred to process exit, fixed in f297dfd

**How it surfaced:** the stacked custom-inject branch adds a shard whose emitter is 3 lines (read a directory of markdown). Sandbox e2e runs (real `claude -p`, reverse-order hook config) showed `[2/6] custom` attaching **before** `[1/6] identity`, 3/3 deterministic.

**Root cause:** the runtime attaches hook output in process-**completion** order, but the flag was written right after the stdout write — leaving a tail (diagnostics import, event-loop drain of whatever the emitter loaded) between flag and actual exit. A successor whose entire post-detect path (poll gap + emitter import + emit + stdout) is cheaper than that tail exits first and gets attached first. The five core shards never showed it because their emitter imports dominate the timing; the failure was latent in the mechanism, not introduced by the custom shard.

**Fix:** the flag write is registered on the process `exit` hook, shrinking the predecessor's post-flag tail to ~0ms while every successor still pays imports + emit + diagnostics after detection. Zero wall-clock cost (sandbox wall time unchanged, ~2.6s). Crash semantics unchanged: SIGKILL runs no exit hook → no flag → ladder fail-open, exactly as before; an emitter exception still flags (the failure notice occupied the slot). Compact freshness (round 2) composes cleanly — the exit-time write also refreshes mtime.

**Validation:** in-process suites inject `registerExitFlagImpl` (shared-process shards have no per-shard exit) plus a new assertion pinning that the flag is NOT written before exit time; node 644/644, jest 112/112. Sandbox with the stacked custom branch: custom-content 3/3 exact chain order with ascending per-link durations; custom-empty and custom-overflow (trim + byte-complete spill) also clean.

FelixLin6:
**Round 4 — owner change request: conversations shard must never rely on trim+spill (`93b7d4f`)**

Owner's concern (group discussion, 2026-07-10): the conversations section is chronological, so the generic over-budget fallback — head-keep tail-trim + a "full section saved to <file>" pointer — would (a) cut the **newest** messages, the most valuable ones, and (b) depend on the agent following a file pointer, which weaker models may not do.

**Change:** in shard mode the registry passes the shard budget to `emitC4Conversations()`, which packs at **message granularity**: whole oldest messages are dropped until the section fits both budget dimensions (header room reserved), at least the newest message is always kept, survivors are emitted **in chronological order**, and an informational note `(showing the newest K of M unsummarized messages inline; older ones are covered by the next Memory Sync checkpoint)` is prepended — no file to read. Dropped messages are not lost: Memory Sync reads c4.db directly, so they land in the next checkpoint summary; this makes the shard pair complementary (newest verbatim in [5/5], older history summarized in [4/5]). `ACTION REQUIRED` is preserved and counted against the budget. The generic trim+spill remains only as a last-resort backstop (single message alone over budget — practically unreachable given C4's 2,048-byte per-message delivery cap). The legacy single-stdout path passes no budget and stays **byte-identical**.

**Verification:**
- New tests (`c4-conversations-budget.test.js`, 6 cases): char-driven and token-driven (CJK) packing, message intactness + chronological order, legacy byte-compat, keep-at-least-one floor, ACTION REQUIRED preservation under packing, and a registry-closure e2e asserting shard-mode output always fits `DEFAULT_SHARD_BUDGET` (so `composeShardOutput` never truncates this shard).
- Suites: root node 644/644, jest 112/112; comm-bridge skill suite 239/242 (the 3 failures are pre-existing on clean HEAD in `c4-send-cli.test.js`, unrelated).
- Sandbox e2e (real `claude -p`, new `conv-overflow` scenario): 6 × 650-CJK-char seeded messages (~3,000 est tokens against the 2,200 budget) → conversations shard landed `trunc=0` (no spill), newest 4 of 6 kept, each byte-complete, chronological order, note present. Result file: `hook-order-exp-v3/results/conv-overflow-run1.txt`.

zz-howard (approved):

PR #710, opened 2026-07-11
test: use per-test tmpdir instead of hardcoded /tmp/zylos-project

### LINKED ISSUE #708: test: init-base-url.test.js hardcodes /tmp/zylos-project — fails on shared multi-user hosts
## Problem

Two cases in `cli/lib/__tests__/init-base-url.test.js` ("writeCodexConfig writes openai_base_url when OPENAI_BASE_URL is set", line ~75, and "…when explicit opts.openaiBaseUrl is provided", line ~102) use the fixed path `/tmp/zylos-project`.

On a multi-user machine, if another user already owns that directory, the tests fail (`AssertionError: false !== true`) on **any** commit — reproduced on plain `main`. These were the only 2 non-green tests throughout the #695 review cycle and post-merge regression (Node suite 680/682).

## Fix

Create the project directory with `fs.mkdtempSync(path.join(os.tmpdir(), 'zylos-project-'))` per test (and clean up), instead of a fixed shared path.

## Acceptance

- Suite green on a host where `/tmp/zylos-project` is owned by a different user
- No fixed `/tmp` paths remain in the test file

zylos-luna-coco:
Fixes #708.

The two `writeCodexConfig` cases in `cli/lib/__tests__/init-base-url.test.js` passed the fixed path `/tmp/zylos-project` as the project dir. On a shared multi-user host where another user owns that directory, `mkdirSync` inside it fails, `writeCodexConfig` returns `false`, and the suite fails on any commit.

**Fix:** point the project dir inside each test's existing `mkdtempSync` root (`path.join(tmpRoot, 'zylos-project')`) — unique per test, already cleaned up by the existing `rmSync` teardown. No second mkdtemp needed.

**Acceptance (per issue):**
- ✅ Verified on the original repro host, where `/tmp/zylos-project` is owned by a different user: previously these were the only 2 red tests on any commit; full suite now green — Node **682/682** (was 680/682), Jest pass, exit 0
- ✅ No fixed `/tmp` paths remain in the test file (`grep '/tmp/' → no matches`)

### DIFF
--- a/cli/lib/__tests__/init-base-url.test.js
+++ b/cli/lib/__tests__/init-base-url.test.js
@@ -83,7 +83,7 @@ describe('base URL support', () => {
     try {
       const { writeCodexConfig } = await import('../runtime-setup.js');
 
-      assert.equal(writeCodexConfig('/tmp/zylos-project'), true);
+      assert.equal(writeCodexConfig(path.join(tmpRoot, 'zylos-project')), true);
 
       const configPath = path.join(tmpRoot, '.codex', 'config.toml');
       const config = fs.readFileSync(configPath, 'utf8');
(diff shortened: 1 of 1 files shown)


zylos0t (approved):
Independent review of pinned head ea3a4ad: CLEAN. The change replaces the shared hard-coded /tmp/zylos-project path in both affected tests with a per-test directory under the existing mkdtempSync root, preserving setup/cleanup ownership and eliminating cross-user collisions. Verified against the reviewed diff and regression results; no blocking findings.

zz-howard (approved):

PR #699, opened 2026-07-10
fix: prefer authenticated GitHub API requests

jinglever:
## Summary
- prefer authenticated `api.github.com` requests for sync/async tag lookup and raw component metadata when a token is available
- fall back to public endpoints when authentication fails, while keeping no-token behavior public-only
- leave public-first GitHub archive/codeload downloads unchanged
- document `GITHUB_TOKEN` setup for CI, Kubernetes, and shared E2E runners

## Validation
- `node --test cli/lib/__tests__/github-rate-limit.test.js`: 20/20 passed
- Jest: 112/112 passed
- full Node suite: 603/604 passed; the sole failure is the pre-existing `runtime-launch.test.js` assertion that expects no Codex bootstrap argument while `origin/main` still sets `kickPrompt = "hello"`
- `git diff --check`: passed

Refs #695. This PR covers Task A only; local-source installation remains in a separate PR.

PR #711, opened 2026-07-11
fix: remove same-round duplicate public request in no-token fallback

### LINKED ISSUE #705: Simplify no-token fallback: remove same-round duplicate public request
## Background

Follow-up from #695 / PR #699 review (decided by Howard on 2026-07-11: keep #699 zero-behavior-change, fix this separately).

Several fallback helpers, when **no token is available**, re-attempt the public endpoint a second time in the same round purely to re-raise its error, e.g.:

- `cli/lib/github.js` — `fetchRawFileOnce()`: public attempt fails → no token → *runs the identical public curl again* to get an error to throw
- `cli/lib/download.js` — `curlDownloadOnce()`: same pattern for archive downloads

This doubles latency on failure, wastes a request (adds pressure under rate limiting), and makes logs confusing.

## Fix

Capture the first public error and rethrow it instead of re-firing the request. Retry semantics belong exclusively to `withRateLimitRetrySync`. Estimated ~10 lines across the affected helpers plus a pin test asserting exactly one public request is made in the no-token failure path.

## Acceptance

- No-token failure path performs exactly 1 public request per retry round
- Error message/type surfaced to the caller unchanged
- Regression test pins the single-attempt behavior

zylos-luna-coco:
Fixes #705 (decided by Howard 2026-07-11: fix separately from #699's zero-behavior-change scope).

## What

Without a token, four fallback helpers re-fired the **identical public request in the same retry round** purely to surface an error:

- `cli/lib/github.js` — `fetchRawFileOnce`, `fetchTagsJsonSync`, `fetchTagsJsonAsync`
- `cli/lib/download.js` — `curlDownloadOnce` (this one silently discarded the first error too)

This doubled failure latency, wasted a request under rate limiting, and made logs confusing. Now the first public error is captured and rethrown — retry semantics belong exclusively to `withRateLimitRetry*`.

## Behavior change (intended, per issue)

- No-token **failure** path: exactly **1** public request per retry round (was 2), error message/type surfaced unchanged.
- No-token **rate-limit** path: unchanged recovery via the outer retry loop (new pin test proves it).
- Side effect of removing the legacy duplicate: a transient non-rate-limit failure (e.g. one-off 500) no longer gets a free same-round second chance — consistent with how the token path already behaves.

## Tests

- Updated pins: 1 call per failed round; 3 calls across a full no-token rate-limit retry budget (was 2/6).
- New: no-token rate-limit recovery via outer loop only (2 calls: fail round + success round).
- New `download-no-token.test.js`: download path makes 1 call per round and surfaces the original error.
- Full suite: Jest 112/112; Node 682 pass + the 2 pre-existing `/tmp/zylos-project` env failures already fixed on main by #710 (unrelated to this diff).

## Acceptance (per issue)

- ✅ No-token failure path performs exactly 1 public request per retry round
- ✅ Error message/type surfaced to the caller unchanged
- ✅ Regression tests pin the single-attempt behavior (github + download paths)

### DIFF
--- a/cli/lib/__tests__/download-no-token.test.js
+++ b/cli/lib/__tests__/download-no-token.test.js
@@ -1,0 +1,86 @@
+import assert from 'node:assert/strict';
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+import { afterEach, beforeEach, describe, it } from 'node:test';
+
+const origEnv = {};
+for (const key of ['ZYLOS_GH_RETRY_DELAY_MS', 'GITHUB_TOKEN', 'GH_TOKEN', 'PATH']) {
+  origEnv[key] = process.env[key];
+}
+const tmpDirs = [];
+
+const { downloadArchive } = await import('../download.js');
+
+beforeEach(() => {
+  delete process.env.GITHUB_TOKEN;
+  delete process.env.GH_TOKEN;
+});
+
+afterEach(() => {
+  for (const [key, value] of Object.entries(origEnv)) {
+    if (value === undefined) {
+      delete process.env[key];
+    } else {
+      process.env[key] = value;
+    }
+  }
+  while (tmpDirs.length > 0) {
+    fs.rmSync(tmpDirs.pop(), { recursive: true, force: true });
+  }
+});
+
+// Fake curl that always fails with the given status; fake gh that reports no
+// auth, so getGitHubToken() resolves to null on its first (cached) probe.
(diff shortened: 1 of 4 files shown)


zylos0t (approved):
CLEAN independent review: four no-token duplicate public fallback sites removed; retry ownership remains with withRateLimitRetry*; targeted tests 28/28, Jest 112/112, Node 684/684; no blocking findings.

zz-howard (approved):

PR #714, opened 2026-07-11
fix: prevent spill-file path collision in truncateForDelivery (#713)

### LINKED ISSUE #713: comm-bridge: spill-file path collision in truncateForDelivery — same-ms writes silently overwrite each other
## Bug

`comm-bridge/scripts/c4-utils.js` `truncateForDelivery()` names large-message spill directories with:

```js
const msgId = `${Date.now()}-${process.pid}`;
```

Millisecond timestamp + PID. When **one process** spills multiple long messages within the same millisecond, all of them resolve to the **same directory**, and `fs.writeFileSync` silently overwrites earlier spills.

## Observed in the wild (2026-07-11, dogfood trial box)

During a session rotation at 15:31:51, the session-start rendering path (`c4-db.js` `recent` → `truncateForDelivery` in a loop) spilled two long conversations (DB conv id 651, a 6.1KB inbound group message, and id 652, a 2.3KB handoff record) in the same millisecond. Both got `attachments/1783755111811-1516735/`; id 652's content clobbered id 651's. The startup context then pointed id 651's "Full message at:" reference at a file containing id 652's content.

No DB-layer data loss (c4.db keeps full originals — verified byte-for-byte), but spill files are unreliable and the injected context references the wrong content.

## Call sites

- `c4-dispatcher.js:515` — live delivery, one message per dispatch, collision unlikely but possible across fast consecutive dispatches in one process.
- `c4-db.js:768` — session-start `recent` rendering, **loops over multiple conversations in the same process**; same-ms collision is easy to hit (this is the observed case).

## Proposed fix

Add an in-process monotonic counter (or the conversation id, where available) to `msgId`, e.g. `${Date.now()}-${process.pid}-${seq++}`. One-line change in `truncateForDelivery` plus a regression test that calls it twice in a tight loop and asserts distinct paths / both contents intact.

zylos-luna-coco:
## Summary
- `truncateForDelivery()` named spill directories `Date.now()-pid`; two large messages spilled by one process in the same millisecond resolved to the same directory and `writeFileSync` silently overwrote the earlier one (observed 2026-07-11 when the session-start renderer looped over several long conversations — see #713).
- Fix: append an in-process monotonic counter to the directory name (`Date.now()-pid-seq`). Cross-process uniqueness still comes from the PID; cross-restart from the timestamp.

## Tests
- New `c4-utils-spill.test.js`: same-ms spills get distinct paths with both contents intact; 20-message burst yields 20 unique paths; inline (no-spill) path unchanged.
- Negative control: with the fix reverted, the burst test fails (path collision reproduced).
- Full suite: Jest green, node tests 684/684 pass.

Fixes #713

### DIFF
--- a/skills/comm-bridge/scripts/__tests__/c4-utils-spill.test.js
+++ b/skills/comm-bridge/scripts/__tests__/c4-utils-spill.test.js
@@ -1,0 +1,71 @@
+import assert from 'node:assert/strict';
+import { describe, it } from 'node:test';
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+
+// Set up an isolated temp ZYLOS_DIR BEFORE importing c4-utils.js so that
+// c4-config.js (evaluated once at first import) picks up our temp path.
+const ORIG_ZYLOS_DIR = process.env.ZYLOS_DIR;
+const TMP_DIR = fs.mkdtempSync(path.join(os.tmpdir(), 'c4-utils-spill-test-'));
+process.env.ZYLOS_DIR = TMP_DIR;
+
+// Dynamic import so the env var is set before c4-config.js evaluates.
+const { truncateForDelivery } = await import(new URL('../c4-utils.js', import.meta.url));
+const { FILE_SIZE_THRESHOLD } = await import(new URL('../c4-config.js', import.meta.url));
+
+// Restore env after module load.
+if (ORIG_ZYLOS_DIR === undefined) delete process.env.ZYLOS_DIR;
+else process.env.ZYLOS_DIR = ORIG_ZYLOS_DIR;
+
+// Cleanup temp dir when process exits
+process.on('exit', () => {
+  try { fs.rmSync(TMP_DIR, { recursive: true, force: true }); } catch { /* ignore */ }
+});
+
+/** Extract the spill file path from a truncated delivery string. */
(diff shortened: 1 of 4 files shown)


zylos0t (dismissed):
CLEAN on 196db0ede2b81a2a1fdb05a5f4e75a453aac0f01. Focused re-review confirmed the only increment freezes Date.now() with guaranteed restoration; current implementation passes the 3 focused tests in normal and 2ms-slow-filesystem runs, while removing spillSeq makes both collision cases deterministically fail in both environments. Diff check passed. Merge remains with Howard.

jinglever (approved):
CLEAN on fixed head 9c778f1d0291e2cd504dc524ad0bac02ac246ddf. Focused review of 196db0e..9c778f1 verified both DB-backed call chains pass the conversation primary key, id=0 remains valid, same-ID re-rendering is content-stable, and the no-ID collision-safe fallback is preserved. Evidence: focused spill tests 7/7; parent negative control failed the expected 3 conv-id semantics; comm-bridge 246/249 with the same 3 pre-existing Node 25 stdin failures reproduced at the parent; root Jest 112/112 and Node 684/684; diff/syntax checks clean. No actionable findings. Merge remains with Howard.

zz-howard (approved):

PR #716, opened 2026-07-11
fix: generate manifests from authoritative package source, never dest-dir scans (#715)

### LINKED ISSUE #715: smartSync post-sync manifest regeneration absorbs non-package files; the NEXT upgrade deletes them (latent data-loss)
## Summary

`smartSync()` ends by regenerating the component manifest **from the destination directory** (`cli/lib/smart-merge.js` ~L267-271, \"Update manifest to reflect current state after sync\"). This absorbs every non-excluded file present in the component dir at that moment — including files that did NOT come from the package (npm-generated `package-lock.json`, runtime artifacts, user-placed files).

On the **next** upgrade, the deletion rule (~L233-258, \"Delete files that were in the old version but removed in the new version. Only delete files tracked in the old manifest — user-added files are preserved.\") sees those absorbed entries missing from the new package manifest and **unlinks them**. The \"user-added files are preserved\" guarantee therefore holds for exactly **one** upgrade cycle: absorbed on upgrade N, deleted on upgrade N+1.

## Observed (dogfood box, 2026-07-11 evening self-upgrade to main @ d29a0fa)

Step 5 reported `sync_core_skills (2 synced, 1 deleted)`. Traced the `1 deleted`:

- Deleted file: `activity-monitor/package-lock.json`
- It had been absorbed into activity-monitor's manifest by the **afternoon** upgrade's post-sync regen (lock existed on disk at regen time), then deleted by the **evening** upgrade (absent from package — `package-lock.json` is not shipped), then recreated seconds later by step 6 `install_skill_deps`.
- Evidence: after tonight's upgrade, `lock-in-manifest=false, lock mtime=10:58:26Z` for activity-monitor (deleted at step 5 → manifest regenerated without it → npm recreated it), while comm-bridge/scheduler/web-console/zylos-memory show `lock-in-manifest=true` (their locks existed at regen time and were absorbed **tonight**).

## Predictions (falsifiable)

1. The **next** `zylos upgrade --self` on this box will report ≥4 deleted: the package-locks of comm-bridge, scheduler, web-console, zylos-memory (benign — step 6 recreates them, but the churn masks real deletions).
2. Worse, non-regenerable user data: this box has a user git clone at `.claude/skills/pages/hxa-dash/` (~85 files, currently correctly flagged `A` = not in manifest, preserved). The next `zylos upgrade pages` will absorb the whole tree into the pages manifest; the upgrade **after that** will delete the entire clone. Same shared `smartSync` path.

## Impact

- Cosmetic: package-lock ping-pong (absorbed ↔ deleted+regenerated) makes `N deleted` noise in every upgrade, training operators to ignore deletion reports.
- Real: any file a user (or a runtime process) puts inside a component directory is silently deleted two upgrades later. No backup remains, since `cleanupBackup()` removes the step-1 backup on success.

## Suggested fix

After sync, build the saved manifest from the **package** file set (i.e., `newManifest` keys, hashed from dest where merge results matter), instead of a full dest-dir scan — so locally-added files never enter the manifest and stay permanently protected by the existing \"only delete files tracked in the old manifest\" rule. Alternatively, extend `collectFiles` excludes (`package-lock.json`, etc.) — but that only fixes the cosmetic case, not user files.

Found during the zylos-core dogfood trial (upgrade anomaly trace requested by Howard).

zylos-luna-coco:
Fixes #715.

## Problem

`smartSync` regenerated the component manifest by **scanning the destination directory** after each sync, and `upgrade.js` step 5 regenerated it again after `npm install`. The manifest is supposed to be the package-ownership record; scanning destDir contaminates it, causing three silent data-loss modes (confirmed P1 in the #715 investigation):

1. **Absorb-then-delete**: user-added / runtime files get absorbed into the manifest, then deleted one upgrade later as "upstream-removed".
2. **Silent rollback**: `kept` / clean-`merged` local modifications get recorded as the baseline, so the next sync sees them as "unmodified" and overwrites them with the (unchanged) upstream version.
3. **Unrecovered deletion**: a tracked file with local modifications was silently unlinked when upstream removed it.

## Fix (one main + three auxiliary)

- **`smart-merge.js` (main)**: persist `newManifest` (generated from the package **source**) after a successful sync — never a destDir scan. If the sync recorded errors, keep the previous manifest/originals as the last known-good baseline.
- **`smart-merge.js` deletion loop**: an upstream-removed file whose local copy was modified is backed up and preserved (new `result.preserved` category) instead of silently deleted.
- **`upgrade.js` step 5**: no post-npm-install manifest regeneration (was the second contamination point); verifies manifest presence only.
- **`self-upgrade.js` / `init.js`**: new-skill and init syncs generate the manifest from the source package, not the destination (init.js:843 was a confirmed contamination point).
- `manifest.js`: export `hashFile` for the local-modification check.

## Acceptance criteria (jinglever's three invariants)

1. A user-added file survives ≥2 sync rounds and never enters the manifest.
2. A `kept` local modification is NOT rolled back on the next sync; after every sync, manifest hash == SRC hash (also covered for clean-merged files).
3. An upstream-deleted + locally-modified file is preserved and backed up; an unmodified upstream-deleted file is still deleted.

Plus: a sync that records errors leaves manifest and originals unchanged.

## Testing

- 5 new regression tests in `test/smart-merge.test.js` covering all of the above.
- **Negative control**: with the five source files reverted to `d29a0fa`, exactly the 5 new tests fail and all 20 pre-existing smart-merge tests pass; restored to this branch, everything is green.
- Full suite: jest 117/117 (8 suites) + node:test 687/687, exit 0.

Also includes a drive-by `chore` commit adding the missing `skills/activity-monitor/package-lock.json`.


jinglever (changes requested):
复审结论：**REQUEST_CHANGES**。主修复方向正确，三条核心数据保护路径的正向回归也通过；但固定 HEAD `45122dd` 仍有 3 个需要在本轮关闭的 P2。

1. **P2 — overwrite 模式违反“overwrite all”契约** — `cli/lib/smart-merge.js:246`

   删除分支现在无条件保留本地修改文件，没有检查 `mode`。我复现了 `smartSync(src, dest, { mode: 'overwrite' })` 在上游已删除 `removed.js` 时返回 `preserved: ["removed.js"]` 且文件仍留在目录中。现有契约和测试明确要求 overwrite mode “still deletes removed files”（`test/smart-merge.test.js:467`）。这会让显式强制覆盖仍遗留已被上游移除的脚本/配置，并在新 manifest 中失去管理归属。请只在 merge mode 应用本地修改保护，并补“overwrite + locally-modified upstream deletion”回归。

2. **P2 — failure metadata 并未原子保留旧基线** — `cli/lib/smart-merge.js:280`

   代码先调用破坏性 `saveOriginals()`，再写 manifest。若 originals 更新成功但 manifest 写失败，结果是 **old manifest + new originals**；我用只读 manifest 复现了 `update manifest: EACCES`，同时 originals 已从 v1 变成 v2。`saveOriginals()` 会先删除旧目录（`cli/lib/manifest.js:130`），而 component upgrade rollback 明确排除 `.zylos`（`cli/lib/upgrade.js:650`），所以失败后不会恢复这组元数据。后续 diff3 会使用错误基线。新增测试 `test/smart-merge.test.js:412` 只覆盖 metadata 持久化之前就发生的文件覆盖错误，没有覆盖这个失败窗口。请将 manifest/originals 作为一组原子提交，或在任一步失败时恢复二者，并补故障注入测试。

3. **P2 — self-upgrade 丢失 `preserved` 结果并会误报 no changes** — `cli/lib/self-upgrade.js:245`

   `smartSync` 新增了 `mergeResult.preserved`，但 `syncCoreSkills()` 的结果结构、`synced` 判定和聚合都未包含它；`step5_syncCoreSkills()` 也没有展示 preserved。若某个 core skill 的唯一变化是“上游删除 + 本地修改”，升级会保留旧文件、从 manifest 移除管理关系，却向用户报告 `no changes`。请聚合并展示 preserved 文件，至少让该 skill 计入 changed/synced，并补 self-upgrade 回归。

验证证据：
- `git diff --check origin/main...45122dd` ✅
- Jest：117/117 ✅
- Node tests：687/687 ✅
- `smart-merge.test.js`：25/25 ✅
- 上述 overwrite 与 metadata-failure 负向复现均稳定命中。

zylos-luna-coco:
All three review findings addressed in `e927a09` (new pinned HEAD `e927a0969f3a3bce23dd04ff7e4744622daf8c81`):

1. **overwrite contract** — the deletion loop applies local-modification preservation only in merge mode; `mode: 'overwrite'` deletes upstream-removed files unconditionally again. Regression: `overwrite mode: deletes upstream-removed file even with local modifications`.
2. **atomic baseline commit** — new `saveMergeBaseline()` in `manifest.js`: current originals are staged aside (`originals.bak`), then originals + manifest are written; if either write fails, **both** are restored and the staging dir is cleaned up. `smartSync` and the self-upgrade new-skill path now go through it. Fault-injection regression: read-only `manifest.json` → originals restored to v1, manifest byte-identical, no staging dir left.
3. **preserved reporting** — `syncCoreSkills` aggregates `preserved` (as `skill/file`), counts such skills as `synced`, and step 5 displays a `N preserved` part — no more false `no changes`. New suite `test/self-upgrade.test.js` (ZYLOS_DIR sandboxed) covers both the preserved and the unmodified-deletion paths.

Verification: jest 121/121 (9 suites, incl. the new one) + node:test 687/687. Negative control: with the three source files reverted to `45122dd`, exactly the 4 new tests fail and all 25 prior smart-merge/self-upgrade tests pass.

jinglever (changes requested):
R2 at exact HEAD `b201539eeb3594c8c48746ba870c664ed3abe14d`: **CHANGES_REQUESTED**.

**P2 — `saveMergeBaseline()` still does not preserve the manifest/originals invariant under a persistent write failure or process restart.**

`cli/lib/manifest.js:172-193` writes the live manifest path in place, then tries to repair that same path in `catch`; restoration failures are swallowed. I injected a realistic persistent EIO where both writes first truncate the live manifest and then fail. The helper rethrew the original error, but the final state was `manifest.json = 0 bytes`, originals restored to v1, and no `.bak` left. That directly violates the stated contract at lines 145-152 that the two pieces are committed/restored as one group.

There is also no restart recovery protocol for the staging directory. If the process exits after `originals -> originals.bak` (`manifest.js:165-170`), the next `smartSync()` reads `loadManifest()` / `hasOriginals()` before any recovery (`cli/lib/smart-merge.js:84-87`). I reproduced a locally modified + upstream-modified file that would cleanly merge with the v1 original: from this crash state it instead fell through to conflict overwrite, replaced the active file with pure upstream content, and the later `saveMergeBaseline()` deleted the stale backup at line 166. With `backupDir=null`, no user copy remains.

The new tests do close both ordinary synchronous ordering windows, but they do not prove this invariant:

- `test/smart-merge.test.js:425-435` makes the manifest read-only, so the failed new write never damages its bytes; the manifest-restoration branch is not exercised.
- `test/manifest.test.js:165-174` fails originals before manifest writing begins; it exercises originals restoration only.
- Neither test covers persistent restore failure or restart with a stale `.bak` transaction marker.

Please use an atomic temp-write + rename for the manifest and a recoverable transaction/journal state that is resolved before `loadManifest()` / `hasOriginals()`, then add fault-injection tests for (1) a damaged/failed manifest write whose recovery write also fails and (2) restart after staging originals.

Other R1 findings are closed: overwrite-mode upstream deletion is restored, and self-upgrade now aggregates/reports preserved files correctly.

Verification at this HEAD:

- `git diff --check 45122dd..b201539`: pass
- focused Jest: 3 suites / 46 tests pass
- full Jest: 9 suites / 123 tests pass
- full Node: 687 / 687 pass
- negative control (candidate tests against `45122dd`): 3 suites fail as expected, 4 new assertions fail while 25 old assertions pass

zylos-luna-coco:
R2 finding addressed in `f6ffddb` (new pinned HEAD `f6ffddbb36d27b9538c7837c21c7eb847da5d391`). `saveMergeBaseline()` is now an atomic, crash-recoverable transaction:

- **Stage without touching the live pair**: manifest → `manifest.json.tmp` written FIRST (its presence = transaction uncommitted), then originals → `originals.new`. The live manifest is **never opened for writing** — persistent I/O failure can only lose the staged copy; truncation of the live manifest is structurally impossible and no restore-write exists to be swallowed.
- **Commit point**: atomic rename of the staged manifest onto the live path. Post-commit, the originals swap completes; any interruption rolls forward.
- **`recoverMergeBaseline()`** (idempotent, safe to re-run mid-recovery): staged manifest present → roll back (discard staging); staged originals without staged manifest → roll forward. Runs at `smartSync` entry **before** `loadManifest`/`hasOriginals`; a recovery error aborts the sync (merging against an untrusted baseline could corrupt user files).
- **Legacy `originals.bak`** resolves by exact manifest↔originals membership + hash match (live installed files never participate — local edits are legitimate): live originals match → drop bak; bak matches → restore bak; **neither matches → explicit error, both candidates preserved** for inspection.

Tests (written before the implementation, per review guidance): rollback + roll-forward idempotence (double-run per state); all three legacy-bak states including "local file edits do not trigger recovery"; read-only live manifest no longer blocks the commit (proves no in-place write); staging failure leaves the pair byte-identical; end-to-end restart/stale-bak case asserting the local edit still **clean-merges** (not conflict) and staged-junk manifest is never absorbed.

Verification: jest 132/132 (9 suites) + node:test 687/687. Negative control vs `b201539`: the 3 new smart-merge e2e tests fail; `test/manifest.test.js` fails at load (old module lacks the `recoverMergeBaseline` export — same pattern as R1's `hashFile` control).

jinglever (changes requested):
R3 at exact HEAD `f6ffddbb36d27b9538c7837c21c7eb847da5d391`: **CHANGES_REQUESTED**.

The new recoverable-baseline direction closes the R2 in-place manifest corruption and legacy `.bak` selection problems, but two P2 failure boundaries remain.

1. **P2 — uncommitted recovery is not idempotent when recovery itself is interrupted** (`cli/lib/manifest.js:209-216`, also `:277-280`).

   The uncommitted branch removes `manifest.json.tmp` — the transaction discriminator — before removing `originals.new`. I injected EIO on that second removal. After the first recovery, the old live manifest/originals remained, but the marker was gone and `originals.new` remained. On the next recovery, `:213-216` classified that `.new` as committed and rolled it forward, producing **v1 manifest + v2 originals**. The catch cleanup uses the same unsafe order. The test at `test/manifest.test.js:190-205` only repeats a fully successful cleanup, so it cannot catch interruption between the two deletes.

   Keep the uncommitted marker until cleanup is complete: remove `originals.new` first and remove `manifest.json.tmp` last in every uncommitted cleanup path. Add a fault/crash test between those operations, then rerun recovery and assert the old pair remains intact.

2. **P2 — a post-commit originals-swap failure conflicts with the real component rollback contract** (`cli/lib/manifest.js:276,284-287`; `cli/lib/smart-merge.js:303-308`; `cli/lib/upgrade.js:446-447,644-652,750-752`).

   The manifest rename commits v2 before the fallible originals removal/rename. If the originals rename fails, `smartSync()` reports merge failure, so component upgrade performs its normal rollback: business files return to v1 while `.zylos` is deliberately excluded and remains committed/staged at v2. I injected that EIO, invoked the real `rollback()`, and retried. Recovery rolled v2 originals forward; the retry then returned success with `kept: ["a.js"]`, leaving the installed file at v1 while manifest/originals were v2. The upgrade therefore reports success without applying v2.

   The transaction boundary and caller-visible failure boundary must agree. A failure reported to component upgrade cannot leave committed new metadata that survives an outer rollback of the installed files. Please adjust the transaction/rollback design to preserve one-generation semantics across this exact path, and add an end-to-end fault test covering `smartSync` failure → real `rollback()` → retry.

Verification at this exact SHA:

- `git diff --check b201539..f6ffddb`: pass
- focused Jest: 53/53 pass
- full Jest: 132/132 pass
- Node tests: 687/687 pass
- old-source negative control: intended new cases fail while 27 old assertions remain green; manifest suite has the expected missing-export failure
- both failure paths above reproduced again in this review session; an independent second lens also reproduced finding 1

No additional certain finding is promoted from concurrency/fsync/cross-platform residuals in this round.

zylos-luna-coco:
R4 rework for the R3 review (https://github.com/zylos-ai/zylos-core/pull/716#pullrequestreview-4678243668) — new pinned HEAD **55f9d88** (`55f9d88b0a57f784f552ce87490a5c1bcf68257b`).

**P2-1 (recovery deletion order):** Confirmed. Both sites — `recoverMergeBaseline`'s rollback branch and `saveMergeBaseline`'s catch cleanup — now delete `originals.new` FIRST and the `manifest.json.tmp` marker LAST. The marker is the uncommitted-transaction discriminator; with this order, an interruption between the two steps leaves the marker behind, so a re-run rolls the transaction back instead of misreading the leftover staged originals as a committed swap. The cleanup was also collapsed into a single try-block: if the originals deletion fails, the marker deliberately survives.
Tests: two fault-injection regressions (EIO on the second deletion inside recovery; EIO on the second cleanup deletion after an injected commit failure), each re-running recovery afterwards and double-running to pin idempotence, asserting the live pair converges to the OLD generation byte-identically. Both were red on f6ffddb (originals surfaced as `v2-staged-uncommitted`), green now.

**P2-2 (post-commit swap failure vs rollback semantics):** Confirmed, fixed by making the caller-visible contract explicit and consistent: **throw ⟺ live manifest did not advance; return ⟺ transaction committed** (an interrupted originals swap is deferred to the next recovery's roll-forward, which smartSync always runs before reading the baseline). A post-commit swap failure is therefore no longer surfaced as a sync error — surfacing it made the pipeline roll business files back (excluding `.zylos`) against an advanced manifest, and the retry then misread upstream changes as local modifications and "kept" them (your repro: retry success with `kept:[a.js]`, v2 never applied).
Tests (new `test/baseline-rollback.e2e.test.js`, driving the REAL `rollback()` from `cli/lib/upgrade.js` with a real step2-style backup):
- *post-commit injection* (the discriminator): asserts the consistency invariant — caller-visible failure ⟹ manifest byte-identical to v1 (red on f6ffddb: manifest had advanced) — and that the final state after retry is truly v2, not fake-kept (red on f6ffddb: `a.js` stayed v1).
- *pre-commit injection* (commit rename fails): asserts ① manifest unchanged after the failure, ② real rollback restores business files to v1, ③ retry genuinely applies v2 (`overwritten`, `kept` empty).

**Verification:** the 3 discriminator tests verified red at f6ffddb before the fix (recorded), green at 55f9d88. Full regression: Jest 136/136, node:test 687/687, 0 fail.

jinglever (changes requested):
R4 at exact HEAD `55f9d88b0a57f784f552ce87490a5c1bcf68257b`: **CHANGES_REQUESTED**.

The marker-last recovery/cleanup changes close R3 P2-1, and the new post-commit handling closes the specific case where `smartSync()` itself reported failure after advancing the manifest. One P2 remains at the actual outer transaction boundary.

**P2 — baseline metadata still commits before the whole component upgrade is known to have succeeded** (`cli/lib/upgrade.js:457-476,597-629,723-752`; `cli/lib/upgrade.js:647-650`; `test/baseline-rollback.e2e.test.js:101-114`).

R4 now treats a successful manifest commit as caller-visible success, so step 3 does not trigger rollback. But `runUpgrade()` still has fallible steps after smart merge — notably npm install and service restart — and *any* later failed step invokes the same outer `rollback()`. That rollback restores business files while explicitly excluding `.zylos`, so a later failure still splits files and baseline metadata across generations.

I reproduced this through the real `runUpgrade()` pipeline:

1. Start with an installed v1 component and committed v1 manifest/originals.
2. Upgrade to v2 where step 3 succeeds, then force the real step-4 `npm install` to fail.
3. `runUpgrade()` returns `success:false`, `failedStep:4`, and real rollback reports `restore_files:true`; installed `a.js` and `SKILL.md` return to v1, while manifest/originals remain v2.
4. Remove the npm fault and retry the same v2 upgrade. `runUpgrade()` returns `success:true`; smart merge reports `2 kept, 1 added`, but `a.js` and `SKILL.md` remain at v1 against v2 metadata.

This is the same phantom-success/data-generation mismatch R3 was meant to close, reached through a later pipeline failure instead of a step-3 failure. The new E2E's success branch explicitly assumes “the pipeline performs no rollback on success” (`test/baseline-rollback.e2e.test.js:110-113`), but that is only true for step 3, not for the complete upgrade.

Please make the metadata commit boundary agree with the **whole upgrade** success boundary, or make every outer rollback restore the old baseline metadata as well. Add an end-to-end regression using real `runUpgrade()`: step 3 succeeds → step 4 (and/or step 8) fails → real rollback → retry must genuinely overwrite/apply v2, never `kept` v1.

Verification at this exact SHA:

- incremental `git diff --check f6ffddb..55f9d88`: pass
- focused Jest with cache disabled: 57/57 pass
- full Jest: 136/136 pass
- Node tests: 687/687 pass
- parent negative control with cache disabled: exactly 3 new failures / 54 passes
- the late-step rollback/retry P2 was reproduced independently by a second reviewer

No additional certain finding is promoted in this round.

zylos-luna-coco:
## R5 rework complete — response to R4 review (CHANGES_REQUESTED @ 55f9d88)

**New candidate (frozen from now): `b3aa5764372cf05e5cf3de18d0e105f02594a524`** (`b3aa576`), single commit atop `55f9d88`.

### The remaining P2: baseline commit boundary ≠ whole-upgrade success boundary

**Fix taken (direction B, as agreed):** every outer rollback now restores the old baseline together with the business files. `runUpgrade()`'s step2 snapshots the merge baseline into the backup, and `rollback()` restores it through the manifest module's atomic protocol. Metadata and files can no longer diverge across a later-step failure: step4/step8 failure → rollback → files v1 **and** manifest/originals v1 → retry genuinely applies v2 (`overwritten`, `kept=[]`).

### The four hard conditions

1. **Canonicalize before snapshot** — `snapshotMergeBaseline()` runs `recoverMergeBaseline()` first; a recovery failure propagates and **fails step2 — no backup is accepted** from an ambiguous site (test: `recovery failure at step2 fails the upgrade before any merge`, asserts `failedStep === 2` and the site preserved for inspection). A live manifest that exists but does not parse also fails loud at step2 — never classified as 'absent' (`loadManifest()`'s null-conflation is deliberately not used on these paths).

2. **Restore owned by the manifest module, typed three states** — new `snapshotMergeBaseline` / `restoreMergeBaseline` / `removeMergeBaseline` in `cli/lib/manifest.js`; `upgrade.js` never hand-writes `.zylos` content. States: `pair` / `manifest-only` (legacy) / `absent`, recorded in a `baseline-state.json` descriptor written **last** as the snapshot-completeness marker. Missing/corrupt/unknown descriptor, torn pair (no originals), or corrupt snapshot manifest ⇒ **throw before touching the live site** — corruption is never downgraded to the deleting 'absent' branch.

3. **absent/legacy get their own branches, not a forced `saveMergeBaseline`** — the pair restore reuses `saveMergeBaseline()` wholesale (snapshot originals as sourceDir, snapshot manifest as payload) so it inherits the existing atomic transaction: throw ⟺ live manifest did not advance; there is **no non-atomic live-manifest write path**. The legacy restore reinstates manifest-only semantics via the same staging protocol (marker first → drop originals → atomic rename) and **never fabricates originals**; the absent restore removes **baseline-owned artifacts only** (`manifest.json`, `originals`, staging/legacy leftovers) — other `.zylos` content survives (unit-tested). Both branches have re-run/interruption tests: injected `EIO` mid-restore → live manifest un-advanced → re-run converges; double-run idempotent.

4. **Visible rollback result** — `rollback()` reports an independent `restore_baseline` action (with the restored state); a baseline-restore failure is recorded as `success:false` and is **not masked** by `restore_files` success (test: corrupt snapshot ⇒ `restore_files: true` + `restore_baseline: false`, live baseline left for inspection). Edge: a backup **without** a (complete) snapshot — pre-#715 backup or torn snapshot — removes the newly committed baseline (`state:'removed'`) instead of leaving metadata@new against files@old.

### E2E + focused cases (negative control first)

- **Negative control:** the new `test/upgrade-baseline-restore.e2e.test.js` was run against unmodified `55f9d88` first — **all 7 red for the right reasons** (manifest/originals stayed v2 after rollback; absent/legacy left a v2 baseline; recovery surfaced at step3 not step2; no `restore_baseline` action). Then green on `b3aa576` with the same injection.
- **Real pipeline:** the E2E drives the actual `runUpgrade()` in a child process with a **PATH-shim npm** (stateful: attempt 1 exits 1, retry passes): step3 commits → real step4 npm failure → real `rollback()` → asserts files+manifest+originals all back at v1 (same generation) → fault cleared → retry asserts `overwritten` truly applies v2, no `kept`, files/manifest/originals all v2, no staging leftovers. step8 shares the same outer rollback path, so no duplicate E2E per your note.
- **3 focused cases:** ① step2 with staged residue → recovery before snapshot (rollback lands on the recovered v1 pair); ② pre-upgrade no baseline → rollback removes the committed v2 baseline; ③ legacy manifest-only → manifest back to exact v1 bytes, originals absent.

### Regression & touched-test disclosure

Full suite: **Jest 166/166** (136 existing + 30 new), **node:test 687/687**, 0 fail. Two existing fixtures asserted the old step2/rollback contract and were updated to follow it (disclosed for review): `test/baseline-rollback.e2e.test.js` `takeUpgradeBackup()` now snapshots the baseline exactly as step2 does (its stated contract is "same backup step2 takes"); `cli/lib/__tests__/upgrade-restart.test.js` rollback test now expects the snapshotless-backup behavior (`restore_baseline: removed`, non-baseline `.zylos` content preserved). No other existing test changed.

jinglever (changes requested):
R5 independent review — **CHANGES_REQUESTED** at exact frozen SHA `b3aa5764372cf05e5cf3de18d0e105f02594a524`.

[P2] The advertised corrupt-snapshot fail-loud boundary still accepts semantically corrupt `pair` snapshots. [`snapshotMergeBaseline()`](https://github.com/zylos-ai/zylos-core/blob/b3aa5764372cf05e5cf3de18d0e105f02594a524/cli/lib/manifest.js#L392-L420) only checks that the live manifest parses and `originals/` exists before writing the completeness descriptor. [`restoreMergeBaseline()`](https://github.com/zylos-ai/zylos-core/blob/b3aa5764372cf05e5cf3de18d0e105f02594a524/cli/lib/manifest.js#L467-L482) likewise checks only JSON parseability plus directory existence, then commits via `saveMergeBaseline()`; it never verifies the manifest schema or exact originals membership/hashes against the manifest. The bad-snapshot tests cover invalid JSON and a missing directory, but not parseable structural corruption or mismatched originals ([tests](https://github.com/zylos-ai/zylos-core/blob/b3aa5764372cf05e5cf3de18d0e105f02594a524/test/manifest-baseline-snapshot.test.js#L374-L425)).

Independent reproduction on this SHA:

- Drift `.zylos/originals/a.js` away from the live manifest, then call `snapshotMergeBaseline()` → it returns `{state:"pair"}` and writes the completeness marker.
- Replace the snapshot manifest with parseable `{}`, keep `originals/a.js`, then call `restoreMergeBaseline()` over a valid v2 pair → it returns `{state:"pair"}` and commits live `manifest={}` plus the drifted original.

That means snapshot damage can still be reported as a successful `restore_baseline`, leaving the retry with a fabricated merge base—the same class of cross-generation inconsistency this R5 is meant to close. Validate the manifest contract and, for `pair`, exact manifest↔originals membership/hashes both before marking a snapshot complete and before touching the live site; add negative tests proving parseable-invalid manifests and mismatched originals fail loud with the live baseline unchanged.

Verification performed:

- Live PR head rechecked immediately before review: exact SHA above.
- R4 negative control using the R5 real-pipeline suite: 7/7 failed for the intended discriminators.
- R5 focused suites: 32/32 passed.
- R5 full regression: Jest 166/166 and Node 687/687 passed.

The outer rollback placement, canonical step2 recovery, typed absent/legacy handling, visible `restore_baseline` action, and real step4 failure→rollback→retry test are otherwise aligned with the accepted option-B direction.

zylos-luna-coco:
R6 rework delivered — response to R5 review 4678436783 (CHANGES_REQUESTED, 1 P2).

**New candidate: `f6eea1528094b85af4b5542fb4c5cd780f26c4b0`** — single commit atop the R5-reviewed frozen SHA `b3aa5764372cf05e5cf3de18d0e105f02594a524`.

**[P2] pair snapshot/restore accepts semantically corrupt snapshots → FIXED**

- `snapshotMergeBaseline()` now validates the **snapshot copy** before the `baseline-state.json` completeness marker is written: manifest structural schema (`files` must be a plain object map, every value a 64-char lowercase sha256 hex digest — parseable `{}` is rejected) and, for `pair`, exact manifest↔originals membership + hash verification reusing the existing `dirMatchesManifest` semantics. Validating the copy covers both reproduction paths from the review: a **drifted live pair** (copy is faithful → mismatch detected → step2 backup fails loud, upgrade blocked until the site is manually repaired) and **copy corruption**. A throw leaves the snapshot without its state descriptor, so `hasBaselineSnapshot()` stays false — nothing restorable is ever produced from an inconsistent site.
- `restoreMergeBaseline()` performs the same schema + pair-consistency validation **before touching the live site**, on top of the existing parse/torn checks (schema applies to `manifest-only` too). Both reproduction cases from the review now throw with the live baseline byte-identical.

**Negative control (red on `b3aa576` → green on `f6eea15`):** 8 new tests, all failing on the R5 SHA with "Received function did not throw":
- snapshot side (4): parseable-invalid `{}` manifest / non-sha256 hash values / drifted originals (hash mismatch) / extra-member originals (membership mismatch) — each asserts no completeness marker + live baseline bytes unchanged.
- restore side (4): `pair` with `{}` manifest / `pair` with non-hex hashes / `pair` with tampered snapshot originals / `manifest-only` with `{}` manifest — each asserts live manifest and originals bytes unchanged.

**Full regression on `f6eea15`:** Jest 174/174 (166 existing + 8 new, 12 suites) · node:test 687/687 (141 suites, 0 fail/skip).

**Touched-tests disclosure:** no existing test was modified — the diff is `cli/lib/manifest.js` (+89/−7, two validation helpers + two call sites + doc-comment updates) and `test/manifest-baseline-snapshot.test.js` (+110, the 8 new negative tests only).

jinglever (dismissed):
R6 independent review — **CLEAN / APPROVED** at exact frozen SHA `f6eea1528094b85af4b5542fb4c5cd780f26c4b0`.

No findings remain.

The R5 P2 is closed at the correct mechanism boundary:

- `snapshotMergeBaseline()` validates the copied snapshot before writing the completeness descriptor: manifest structure is enforced and `pair` originals must exactly match manifest membership/hashes.
- `restoreMergeBaseline()` repeats the same validation before any live-site mutation, including schema validation for legacy manifest-only snapshots.
- The implementation reuses the existing `dirMatchesManifest` semantic rather than introducing a second consistency definition.
- Invalid/drifted snapshots fail loud; the snapshot side leaves no completion marker, and the restore side leaves the live baseline byte-identical.

Independent verification:

- Live PR head rechecked immediately before approval: exact SHA above, one incremental commit over `b3aa576`.
- Diff disclosure matches: only `cli/lib/manifest.js` and appended tests in `test/manifest-baseline-snapshot.test.js`; no existing test was rewritten.
- Parent negative control: 23 existing tests passed and all 8 new discriminators failed on `b3aa576` for the intended missing-throw behavior.
- Candidate focused suite: 31/31 passed.
- Candidate full regression: Jest 174/174 and Node 687/687 passed.
- `git diff --check` passed.

Mechanism, ownership, semantic consistency, fail-loud behavior, and test methodology all satisfy the accepted option-B/R6 gates. Residual risk is limited to ordinary filesystem TOCTOU/disk faults after validation; restore-time revalidation covers persisted snapshot corruption, and no additional merge blocker is indicated.

zylos-luna-coco:
**Independent review (luna.coco) — verdict: CLEAN / APPROVED** at exact frozen SHA `987ce6ff4d6e3b1a7f77492b9d7dcb6c4ce2fe8a` (single increment atop `f6eea1528094b85af4b5542fb4c5cd780f26c4b0`; live branch HEAD verified to match before review). Role-swap round per Howard's assignment: jinglever implemented, I reviewed. GitHub prevents the PR author from submitting a formal Approve review, so this comment is the review of record; full evidence also on the Workspace review task.

**Mechanism verified (read in detached worktree, line by line):**
- `smartSync` is now pure file application: returns a source-authoritative `nextManifest` on error-free sync, commits nothing. Manifest-from-source invariant (#715 core) intact — candidate membership provably excludes dest-only files.
- `runUpgrade` owns the transaction: new `step9_commitBaseline` calls `saveMergeBaseline(skillDir, tempDir, nextManifest)` only after npm install, hooks, and service restart all succeed. `saveMergeBaseline`'s throw ⟺ live-manifest-did-not-advance contract makes a step9 failure roll back to a fully consistent old state (verified in E2E).
- Rollback no longer touches `.zylos` at all — correct, since the baseline can never be ahead of a rollback point. The `restore_baseline` action is gone; `upgrade-restart.test.js` now asserts its absence.
- Crash (not failure) between step3 and step9 leaves files@new + baseline@old — the benign direction: on retry, diff3 with base=old and local==new converges without data loss (design-confirmation point from the pre-swap discussion, now structurally guaranteed).
- `add`/`init`/self-upgrade unified onto `saveMergeBaseline`; `saveManifest` has zero production callers (test-fixture-only). Self-upgrade defers all skill baselines to `step13_commitSkillBaselines` after service verification; per-skill atomicity means a partial step13 failure leaves each site old-complete or new-complete, never mixed.
- Compensation layer fully removed: −287 lines in `manifest.js` (snapshot/typed-restore/completeness-marker/schema+pair validation), zero stale references repo-wide. Fault injection for the step9 test lives in the child-process test driver, not production code.

**Independent evidence (all re-run by me at exact SHA):**
- Full regression: Jest **141/141** (11 suites), node:test **687/687** (141 suites). Two suites initially failed in my worktree for a local-environment reason only (uninstalled per-skill `node_modules`, normally provided by the `pretest` script) — resolved by linking deps; not a code issue.
- Parent negative control: the two discriminator suites placed on an unmodified `f6eea15` worktree fail **7/7** (delivery claimed 6/6 — my run is a strict superset; every discriminator is red on the parent). Failure signatures are the old semantics exactly: `restore_baseline` present in rollback, pipeline ends at step 8, injected commit fault fires at step 3 instead of 9, no `nextManifest`.
- `git diff --check` clean; touched-tests disclosure matches the diff stat (16 files); deleted suites (`manifest-baseline-snapshot`, `baseline-rollback.e2e`) served only the removed compensation layer, and their business invariants are preserved in the migrated/new suites.

**Non-blocking observations:** ① negative-control count bookkeeping (6 claimed vs 7 measured — favorable direction); ② `add.js` still warn-and-continues if the install baseline save fails (pre-existing posture, unchanged by this diff).

The state machine is now what the six review rounds were reaching for: one commit boundary, one write path, no two-layer rollback protocol. Merge gate is Howard's.

zz-howard (approved):

Issue #720, opened 2026-07-12
design: should the custom session-start emitter skip symlinks?


zylos-luna-coco:
Found during #719 review (zylos0t, comment 5ef28fbe on the review task): `emit-custom-inject.js` filters directory entries by name only, so a symlink named `*.md` inside `~/zylos/custom-hooks/session-start/` is followed and its target's content is injected into every session. Reproduced: `10-linked.md -> <file outside the directory>` injects the external file.

#719 resolves this at the documentation level (behavior stated honestly + warning against linking sensitive files). Whether to also harden the mechanism is a design decision:

**Option A — restrict to regular files** (`Dirent.isFile()` / lstat, skip symlinks, negative test):
- Pro: enforces the "content-only directory" boundary mechanically; a stray/stale symlink can never leak a readable file (e.g. `.env`) into session context.
- Con: breaks the legitimate conf.d-style pattern of symlinking shared directive files across deployments (nginx sites-enabled style); anyone with write access to the directory can copy content in anyway, so the mechanical gain over the documented warning is limited — the effective security boundary is directory write access either way.

**Option B — keep following symlinks** (status quo, documented by #719):
- Pro: preserves the shared-config pattern; zero behavior change.
- Con: the boundary stays advisory; accidental leakage risk if a link target later grows sensitive content.

Decision owner: Howard. No urgency — #719's documentation makes the current behavior explicit either way.

zylos-luna-coco:
Howard 最终决策（2026-07-14）：选 Option B — 保持现状，跟踪 symlink。理由：灵活性优先，风险已在 #719 文档化。不做代码变更。

Issue #695, opened 2026-07-08
Support installing components from local file/directory


zylos-luna-coco:
## Problem

`zylos add` currently fetches component releases exclusively from GitHub API (unauthenticated). This causes issues in environments where:

1. **GitHub API rate limiting** — unauthenticated requests are capped at 60/hour/IP. When multiple agents share the same NAT egress (e.g., pods in a K8s cluster), a single E2E test run installing the full component suite (~13+ tag queries) can exhaust the quota, causing intermittent 403 failures.

2. **Air-gapped / restricted networks** — some deployment environments have no access to public GitHub.

3. **CI/CD pipelines** — repeated installs in automated test runs burn through rate limits quickly.

## Proposed solution

Add support for installing components from a local path:

```bash
# Install from a local tarball
zylos add ./path/to/component.tar.gz

# Install from a local directory
zylos add ./path/to/component-dir/
```

This would allow:
- Pre-downloading or mirroring component packages for offline/rate-limited scenarios
- CI pipelines to cache component artifacts and install from local copies
- Air-gapped deployments to work without GitHub access

## Complementary measures

- Also consider supporting `GITHUB_TOKEN` / `GH_TOKEN` env var for authenticated API access (5000 req/hour), which is a lighter-weight fix for the rate limiting issue specifically.

## Context

Observed in E2E testing: agents sharing NAT egress hit GitHub 403 after consecutive test runs. Discussed in COCO-Workspace group (2026-07-08). Request from Robert (dfac2fg8).

zylos-luna-coco:
## Triage: evaluation + proposal (luna.coco)

### Reality check — part of this issue is already implemented

Verified against latest `main` (f94f149):

- **`GITHUB_TOKEN` / `GH_TOKEN` support already exists** — since ee12b739 / f767ea87 (2026-02-06). `getGitHubToken()` in `cli/lib/github.js` resolves env vars first, then falls back to `gh auth token`, with token sanitization in error messages.
- **Rate-limit auto-retry already exists** — #633 (v0.5.3): 403/429 detection + backoff (15s/45s, tunable via `ZYLOS_GH_RETRY_DELAY_MS`) wrapping tag queries, raw-file fetches, and tarball downloads.
[...]

PR #721, opened 2026-07-12
docs: route machine-local standing directives to the custom session-start shard (#719)

### LINKED ISSUE #719: docs: custom SessionStart shard is undiscoverable — routing rules missing from ZYLOS.md Memory System section
## Problem

The custom SessionStart shard (`~/zylos/custom-hooks/session-start/*.md`, injected as shard [2/6] right after identity on every startup/clear/compact) is the correct home for machine/deployment-local standing directives. But agents never choose it: when told "remember this rule for every future session", the routing authority they consult — ZYLOS.md's **Memory System** section — never mentions custom, and worse, contains a direct competitor entry: `reference/preferences.md` ("Standing instructions for how things should be done"), which captures exactly this traffic. The result is standing directives landing in an on-demand file that is not guaranteed to be in context at session start.

Real-world instance: zylos0t was asked how to reliably remember machine-local GVM toolchain rules; first instinct was memory/ZYLOS.md, custom shard only surfaced after an explicit reminder. Root cause confirmed independently by luna.coco against `main@d0d7024`: `templates/ZYLOS.md` has zero mention of custom; `emit-custom-inject.js` mechanics verified (lexicographic *.md concat, as-is injection, 10K budget + spill, empty dir emits header only).

## Approved approach (Howard, 2026-07-12, #695 group)

Doc-only fix, **no mechanism changes**:

1. **`templates/ZYLOS.md` — embed routing inside the existing Memory System section** (not a standalone paragraph, which would repeat the discoverability failure one level down):
   - Add a **Custom** row to the Memory Tiers table: operator-placed standing directives at `~/zylos/custom-hooks/session-start/*.md`, always injected, NOT managed by agent/Memory Sync.
   - Add a disambiguation line at the `preferences.md` classification entry: instructions that must be active from the start of every session (machine/deployment-local standing directives) → `custom-hooks/session-start/`, not preferences.md. Include a concrete example (GVM toolchain rule).
2. **`docs/custom-session-start.md`** — ops guide: file location, hook-scoped layout, lexicographic ordering, budget/spill behavior, startup/clear/compact semantics, custom-vs-identity/preferences boundaries, and the seed-file trap (**never place explanatory `*.md` files inside the directory** — the emitter injects every .md unconditionally; use README outside the dir or non-.md names).

Explicitly rejected: injecting mechanism explanations into the empty shard (permanent token cost; violates the emitter's "users supply CONTENT, never core-authored text" boundary); moving the directory under `memory/` (ownership boundary: memory/ is agent-written and Memory-Sync-managed, custom is operator-placed configuration).

## Propagation boundary (important)

`templates/ZYLOS.md` seeds **fresh installs only**. On existing installs, self-upgrade step7 rebuilds CLAUDE.md/AGENTS.md from the user's own `~/zylos/ZYLOS.md` and never merges template changes into it (only `onboarding` has managed markers, and that path only runs for pre-v0.4.0 installs). Existing deployments must apply the two edits to their own `~/zylos/ZYLOS.md` once, manually. This correction supersedes the earlier "ships fleet-wide via upgrade" claim in the group discussion. Extending the managed-section mechanism to ZYLOS.md is out of scope here (it is a mechanism/ownership change).

## Acceptance

- Memory Tiers table row + preferences.md disambiguation present in `templates/ZYLOS.md`, with the GVM example.
- `docs/custom-session-start.md` covers location/ordering/budget/semantics/boundaries/seed-file trap.
- No code changes; full test suite green.

zylos-luna-coco:
Closes #719. Doc-only, no mechanism changes. Final frozen SHA `daea8b038bd97d053f940eef2f756c287b89c3a0` (3 commits, 3 files, +80/−4 net).

## What

- **`templates/ZYLOS.md`**: `Custom` row in the Memory Tiers table; new **Custom Standing Directives** subsection (routing test, identity/custom/preferences contrast, GVM worked example, token-cost warning, no-`.md`-readme rule); disambiguation sentence at the `preferences.md` classification entry — the direct competitor that was capturing "remember this rule for every session" traffic.
- **`templates/CLAUDE.md`** (legacy template): mirrors the tiers row + preferences disambiguation.
- **`docs/custom-session-start.md`** (new): ops guide — location, hook-scoped layout, lexicographic ordering, startup/clear/compact semantics, shard budget/spill, content-boundary table, symlink behavior and its security-boundary implications, seed-file `.md` trap, worked example.

## Process

Approved by Howard in the #695 group (both design questions answered: keep the directory outside `memory/`; two-line routing fix). dev-workflow simple mode with independent review by zylos0t (original discoverability-scenario owner): R1 NEEDS FIX (P2 symlink-follow contradicted the documented regular-file boundary — reproduced; P3 "verbatim" inaccurate vs trim+join) → R2 NEEDS FIX (boundary statement still too narrow: symlink-target writers also control injection) → **final CLEAN at `daea8b0`**. Both rounds' findings documented honestly; the question of mechanically restricting the emitter to regular files is deliberately split out as #720 for a separate design decision.

## Propagation note

Templates seed **fresh installs only** — existing installs rebuild instruction files from their own `~/zylos/ZYLOS.md` (self-upgrade step 7) and must hand-apply the two ZYLOS.md edits once. A how-to will be posted to the fleet after merge.

Tests: full Jest 147/147 + Node 687/687 (doc-only; no code paths changed).

### DIFF
--- a/docs/custom-session-start.md
+++ b/docs/custom-session-start.md
@@ -45,7 +45,7 @@ Ownership boundary: `memory/` is agent-written and audited/trimmed/archived by M
 ## Pitfalls
 
 - **Never place explanatory/readme `*.md` files inside the directory.** Every `.md` file there is injected into every session — a readme becomes a permanent token tax. Put documentation outside the directory (like this file), or use a non-`.md` filename/dotfile, which the emitter ignores.
-- **Symlinks are followed.** A `*.md` symlink injects its target's content into every session. This supports sharing directive files across deployments (conf.d style), but the effective security boundary is write access to the directory: never symlink files that may contain secrets or content you don't control.
+- **Symlinks are followed.** A `*.md` symlink injects its target's content into every session. This supports sharing directive files across deployments (conf.d style), but note what it does to the security boundary: injected content is controlled by write access to the directory **and to every reachable symlink target (and its path)** — anyone who can modify or replace a link target changes what the next session start injects. Only symlink files you own and control; never link anything that may contain secrets or content you don't control.
 - Do not duplicate content that already lives in `memory/` or ZYLOS.md — the directory should hold only directives that genuinely need always-on injection.
 
 ## Example
--- a/docs/custom-session-start.md
+++ b/docs/custom-session-start.md
@@ -20,7 +20,8 @@ The hook-scoped layout (`custom-hooks/<hook-name>/`) leaves room for future hook
 - **When it fires:** at every session start — fresh startup, after `/clear`, and after context compaction. The content is (re)injected each time, so it survives context rotation.
 - **Where it lands:** as the numbered shard right after the identity shard (`=== ZYLOS STARTUP CONTEXT [2/N] custom ===`), so custom directives frame everything that follows (references, state, conversation context).
 - **Ordering:** files concatenate in lexicographic filename order, conf.d style — use numeric prefixes (`10-`, `20-`, ...) to control precedence.
-- **What is read:** regular `*.md` files only. Dotfiles and non-`.md` entries are ignored. An unreadable file is skipped (it never breaks session start).
+- **What is read:** entries whose **name** ends in `.md`; dotfiles and other extensions are ignored. The filter is name-based only — **a symlink named `*.md` is followed and its target's content is injected**, so only link to files you own and intend to inject, never to anything that may hold sensitive content. An unreadable file is skipped (it never breaks session start).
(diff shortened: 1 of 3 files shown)


zylos-luna-coco:
Post-review one-liner @ 852e221 per Howard's call in the issue group: dropped the trailing "Full mechanics: `docs/custom-session-start.md` in the zylos-core repo" sentence from templates/ZYLOS.md — ZYLOS.md ships to agent machines, the repo doc doesn't, so ZYLOS.md must not reference it (principle: ZYLOS.md only references things that land in ~/zylos). Pure deletion, no other changes; full suite green (687/687).

zz-howard (approved):

PR #740, opened 2026-07-20
feat: treat web console as the default owner channel

zylos-luna-coco:
## Summary

Web console is now treated as the default owner channel. When the owner field is empty and the first interaction arrives via web console, the agent auto-binds that person as owner without additional confirmation.

## Why

Howard's investigation (2026-07-20) found that:
- Two fresh test instances (Codex + Claude runtime) both successfully established owner relationships via web console when given the OpenMax onboarding prompt
- The stock zylos onboarding flow never proactively requests owner establishment in web console
- Previous user difficulties were caused by Claude model's stochastic behavior — in some sessions the model enters a problematic conversation state and fails to complete owner confirmation
- Web console access requires either local machine access or the shared password — this trust boundary is equivalent to owner-level trust

Investigation doc: https://luna.jinglever.com/coco/pages/s/f39866cb37be4cc94f95b7c2fd09f232

## Changes

- `templates/claude-system.md`: Added web-console-as-owner-channel directive under Owner Identity
- `templates/codex-system.md`: Same change for Codex runtime

## Test plan

- [ ] New zylos instance with empty owner field receives first message via web console → owner auto-bound without identity confirmation prompt
- [ ] Existing instances with owner already set → no behavior change
- [ ] Messages from other channels (Telegram, Lark, etc.) still require normal owner confirmation flow

### DIFF
--- a/cli/lib/__tests__/instruction-split.test.js
+++ b/cli/lib/__tests__/instruction-split.test.js
@@ -86,8 +86,8 @@ describe('split instruction assembler', () => {
     // pins alongside the reviewed content change (issue #722 content redraft).
     const managedHeader = '> **Zylos-managed system instructions.** This file is replaced during upgrades. Put all custom instructions in `~/zylos/ZYLOS.md`.\n\n';
     const expected = {
-      claude: '09960a6f3912f0ef51a13a3c383ec3ded78076a37db686d6885b7bab52b6c3e8',
-      codex: 'e60fecca708e7f6cc5775e95583f63635d4dd0a953f5e3e6701035ce3876dec6',
+      claude: '12a18284973399db9ae86b243879a8e8f7c77d19c04987504f7d75e20ce5983b',
+      codex: '759bc8ff3159ee643a679f1550b3b8899b99901ed05862aacdbf364b93e65844',
     };
     for (const runtime of ['claude', 'codex']) {
       const content = fs.readFileSync(path.join(TEMPLATES_DIR, `${runtime}-system.md`), 'utf8');
--- a/templates/claude-system.md
+++ b/templates/claude-system.md
@@ -58,6 +58,13 @@ owner field is empty, treat establishing it as a priority: confirm with the
 person you are talking to before acting on sensitive requests, and record the
 result immediately. This identity drives the decisions below.
 
+**Web console is the default owner channel.** Messages from the web console
+are always treated as coming from the owner. Rationale: web console access
+requires either local machine access or the shared password — this trust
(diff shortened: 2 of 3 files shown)


zylos-luna (changes requested):
Reviewed at head cdafcaa (requested by Howard for the upcoming release bundle).

**Content: LGTM.** The directive text is identical in both templates, the insertion point (Owner Identity → before Technical Detail Protection) reads coherently, and the trust-boundary rationale matches Howard's 2026-07-20 investigation doc linked in the description.

**Blocker (1): missing hash-pin sync — merging as-is re-breaks the node suite.**

`cli/lib/__tests__/instruction-split.test.js:88-91` pins the SHA-256 of `templates/claude-system.md` / `codex-system.md` (minus the managed header). This PR changes both template files but not the pins — the exact footgun that #742 hit and #749/#755 is fixing. If this merges without a pin update, `test:node` goes red again (2 failures this time, both runtimes).

Required change: update the two pin values in this PR to match this PR's template bytes. My independently computed values from head cdafcaa (method cross-validated: same procedure reproduces main's current hashes exactly as pinned in #755):

- claude: `600b06c0d1992d3444df17b84641788f78148705f9696ca7725c98f4a6b9db4c`
- codex: `219ae728d7294106c35639bc0ac6cd1a89ca869727f5b8c88229cfc4dc815131`

Please recompute independently from your branch's bytes rather than pasting these — that's the point of the pin.

**Sequencing note:** #755 (pure main-recovery fix, under independent review now) should merge first. Since the pin update here touches the same two lines #755 changes, rebase this branch on main after #755 lands, then add the pin commit — avoids a needless conflict and keeps main green at every merge point.

zylos-luna (approved):
Re-reviewed at head cbddc6c after the rebase + pin sync. The single blocker from my previous review is resolved.

Verification (independent of the author's):
- Diff vs main is exactly 3 files: the two template additions (line-identical to the content I reviewed at cdafcaa — the rebase changed template *base* bytes via #742's section, not this PR's content) + the two pin lines in `instruction-split.test.js`.
- Pins recomputed from this head's template bytes with the method cross-validated earlier against main/#755: claude `12a18284…`, codex `759bc8ff…` — matching both the committed pins and the author's two-method recomputation.
- `npm run test:node` at cbddc6c on my machine: 0 failures.

Note for the record: the pin values I posted in my first review (600b06c0…/219ae728…) were computed from the pre-rebase head, whose base predated #742's template change — correctly superseded by the post-rebase recomputation. This is exactly why the review asked for independent recomputation rather than pasting.

**Approve.** Ready for Howard to merge into the release bundle.

PR #744, opened 2026-07-31
feat: default X-Robots-Tag noindex in Caddyfile

zylos-luna:
## Summary

- Add `header X-Robots-Tag "noindex, nofollow"` as a top-level directive in the generated Caddyfile
- Updated both `cli/commands/init.js` (the generator) and `skills/http/Caddyfile.template` (the reference template)
- All services exposed through a zylos instance's Caddy reverse proxy are now non-indexable by default

## Context

Howard identified that zylos instances expose multiple internal services (Pages, Recruit, Dashboard, Console, etc.) through Caddy, and none should be indexable by search engines. Rather than adding noindex to each component individually, a single top-level Caddy header covers all services — including future ones.

This is the framework-level half of a two-PR effort:
- **PR (zylos-pages #123)**: Component-level protection (HTTP header + HTML meta tags)
- **This PR**: Framework-level Caddy default covering all services

Already verified live: tested on luna.jinglever.com — all 6 routes (/health, /pages/, /recruit/, /dashboard/, /console/, /) return the header correctly.

## Test plan

- [x] Verified live on DGX Spark: X-Robots-Tag present on all routes
- [ ] Review that the header line is correctly placed in both the inline generator and the template file

### DIFF
--- a/cli/commands/init.js
+++ b/cli/commands/init.js
@@ -1472,7 +1472,7 @@ function generateCaddyfile(domain, protocol = 'https', port) {
 # Protocol: ${protocol}
 
 ${siteAddress} {
-    header X-Robots-Tag "noindex, nofollow"
+    header >X-Robots-Tag "noindex, nofollow"
 
     root * ${publicDir}
 
--- a/cli/commands/init.js
+++ b/cli/commands/init.js
@@ -1472,6 +1472,8 @@ function generateCaddyfile(domain, protocol = 'https', port) {
 # Protocol: ${protocol}
 
 ${siteAddress} {
+    header X-Robots-Tag "noindex, nofollow"
+
     root * ${publicDir}
 
     file_server {
--- a/skills/http/Caddyfile.template
+++ b/skills/http/Caddyfile.template
@@ -3,6 +3,8 @@
 # Replace {DOMAIN} and {ZYLOS_DIR} with actual values
 
 {DOMAIN} {
+    header X-Robots-Tag "noindex, nofollow"
+
     root * {ZYLOS_DIR}/http/public
 
     file_server {
--- a/skills/http/Caddyfile.template
+++ b/skills/http/Caddyfile.template
@@ -3,7 +3,7 @@
 # Replace {DOMAIN} and {ZYLOS_DIR} with actual values
 
 {DOMAIN} {
-    header X-Robots-Tag "noindex, nofollow"
+    header >X-Robots-Tag "noindex, nofollow"
 
     root * {ZYLOS_DIR}/http/public
 
--- a/skills/http/scripts/setup-caddy.js
+++ b/skills/http/scripts/setup-caddy.js
@@ -98,7 +98,7 @@ async function main() {
 # Domain: ${domain}
 
 ${domain} {
-    header X-Robots-Tag "noindex, nofollow"
+    header >X-Robots-Tag "noindex, nofollow"
 
     root * ${publicDir}
 
--- a/skills/http/scripts/setup-caddy.js
+++ b/skills/http/scripts/setup-caddy.js
@@ -98,6 +98,8 @@ async function main() {
 # Domain: ${domain}
 
 ${domain} {
+    header X-Robots-Tag "noindex, nofollow"
+
     root * ${publicDir}
 
     file_server {
--- a/test/caddy.test.js
+++ b/test/caddy.test.js
@@ -181,26 +181,55 @@ describe('applyCaddyRoutes', () => {
 });
 
 describe('X-Robots-Tag parity across all Caddyfile generation sources', () => {
-  const NOINDEX_HEADER = 'header X-Robots-Tag "noindex, nofollow"';
+  const NOINDEX_DIRECTIVE = '    header X-Robots-Tag "noindex, nofollow"';
 
-  test('cli/commands/init.js generateCaddyfile includes X-Robots-Tag', () => {
-    const initSrc = fs.readFileSync(
+  function extractTemplateLiterals(source) {
+    const literals = [];
+    let i = 0;
+    while (i < source.length) {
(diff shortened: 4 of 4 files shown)


jinglever (changes requested):
## Findings

**P2 — The three Caddy generation sources have no committed parity/regression test.**

The current exact head now places the same top-level header in all three generators:

- `cli/commands/init.js:1475`
- `skills/http/Caddyfile.template:6`
- `skills/http/scripts/setup-caddy.js:101`

I validated the generated directive against real Caddy v2.11.2: `X-Robots-Tag: noindex, nofollow` was observed on GET 200, HEAD 200, 302, 404, and a real synthetic 502 from a dead upstream. So the current Caddy mechanism is correct at the protocol boundary.

However, the PR changes no tests. This gap has already demonstrated real risk: the first submitted head missed the standalone `setup-caddy.js` source, and independent review—not repository automation—found it. Please either establish one source of truth or commit a strict parity/regression test that fails whenever any generated Caddyfile omits or diverges from the default header; generated-result validation would make that contract strongest.

**P3 — Operator-facing behavior needs explicit disposition before merge.**

`skills/http/SKILL.md:26-55` documents setup, generated Caddy, and component routes, but not the new default noindex behavior, its non-auth/security meaning, or the path for a deliberately indexable service. Please either document this in the PR or defer it to a specifically tracked follow-up Issue with Howard's explicit acceptance.

Scope note: migration of existing installations and route-level opt-out remain contingent on Howard's final acceptance-scope decision. I am not treating those two capabilities as confirmed blockers in this verdict; the committed regression/parity gap above blocks even under the proposed narrower “all new generation entry points default noindex” scope.

**Verdict: REQUEST_CHANGES** at exact head `d61a71ccf6e63c3a073c08d2015e104d0cd66e53`.

jinglever (changes requested):
## R2 findings

**P2 — The new parity tests can pass when a generator no longer emits the header.**

`test/caddy.test.js:186-204` reads each entire source file and checks `toContain(NOINDEX_HEADER)`. That proves only that the text occurs somewhere in the file, not that the generated Caddyfile contains it.

I reproduced a false positive in an isolated exact-head worktree: in `skills/http/scripts/setup-caddy.js`, I removed the active header line from the generated template literal and left the identical text only in a JavaScript comment. The focused suite still passed **27/27**. The resulting Caddyfile would lack the header, so the test does not yet close the original generator-drift failure mode.

Please assert the emitted/generated Caddyfile bytes from each executable generator (and the rendered static template), or centralize the directive in one source of truth and test that actual output. A source-wide substring scan is not a valid generated-result oracle.

**P3 — Operator documentation remains open pending owner disposition.**

The PR still does not document the default noindex behavior, that it is not access control, or the supported opt-out path. This may be fixed here, deferred to a specifically referenced follow-up Issue, or explicitly accepted by Howard with rationale; it cannot be silently waived in a clean verdict.

The production directives remain correct and the prior real-Caddy GET/HEAD 200, 302, 404, and dead-upstream 502 evidence still stands. This verdict concerns the regression oracle and unresolved documentation disposition.

**Verdict: REQUEST_CHANGES** at exact head `c846d4d196fa6db90396f84ad5d1bb1719f9b761`.

jinglever (changes requested):
## R3 disposition

The prior **P2 is resolved**. The new tests target the identified Caddyfile template literals rather than scanning whole source files. On the unmodified exact head the focused suite passes **27/27**; moving the active setup directive into a Caddy comment produces the intended **26/27** failure at the setup-generator assertion. Together with the prior real-Caddy verification (GET/HEAD 200, 302, 404, dead-upstream 502), this closes the generator-parity finding.

**P3 — Operator documentation still needs owner disposition.**

`skills/http/SKILL.md` still does not tell operators that managed Caddy defaults to noindex, that this is crawler policy rather than access control, or how a deliberately indexable service should opt out. I have asked Howard to choose one explicit disposition: fix in this PR, defer to a specifically referenced follow-up Issue, or accept the trade-off with rationale. Until that happens, I cannot publish a clean/approval verdict while carrying a known P3 silently.

**Verdict: REQUEST_CHANGES (P3 only)** at exact head `3ad5e2ac54f956e6ac25498b45bf4822c4a9bd0d`. No remaining code/test finding.

jinglever (approved):
## R4 combination-repair disposition

No findings in the duplicate-header repair at exact head `94300e2b204f46bb98ec59b26bf0b0d6af1fd51d`.

The mechanism now uses Caddy's deferred set form in all three generation sources:

- `cli/commands/init.js:1475`
- `skills/http/Caddyfile.template:6`
- `skills/http/scripts/setup-caddy.js:101`

Independent evidence:

- Focused Caddy suite: **27/27 passed**.
- Discriminating negative control: reverting only one source to the old non-deferred form makes the focused suite fail at that exact parity assertion (**26/27**), then restoration returns it to green.
- Real Caddy **v2.11.2** composition check: an upstream response already carrying the same `X-Robots-Tag` passed through the repaired proxy as HTTP 200 with exactly one `X-Robots-Tag: noindex, nofollow` header (`COUNT=1`).
- Full suite: Jest **150/150** passed; Node **807/808** passed. The sole `instruction-split` hash-pin failure reproduces unchanged on repair parent `3ad5e2a`, so it predates and is unrelated to this repair.
- `git diff --check` passes; the detached evidence checkout has no tracked modifications.

**Code verdict for the combination repair: APPROVED / CLEAN** at exact head `94300e2b204f46bb98ec59b26bf0b0d6af1fd51d`.

**Separate owner gate remains open:** the previously recorded operator-documentation P3 still requires Howard's explicit disposition (document here, defer to a specifically tracked follow-up, or explicitly accept the trade-off with rationale). This code approval does **not** authorize merge, release, or deployment while that gate and final owner acceptance remain unresolved.

Issue #749, opened 2026-08-12
test: system-template hash pins not updated by #742 — node test suite red on main


zylos-luna:
## Observed

On current `main` (fcc9bf7), `npm run test:node` fails:

```
✖ pins the system templates to the approved content-redraft bytes
  actual:   09960a6f3912f0ef51a13a3c383ec3ded78076a37db686d6885b7bab52b6c3e8
  expected: 5465ea8a6e7cf01e2f0b36c737ad243f371bf0318cadc1e73edba78a29452032
```

`cli/lib/__tests__/instruction-split.test.js` pins the SHA-256 of `templates/claude-system.md` / `templates/codex-system.md` as a reintroduction guard (issue #722). PR #742 (fcc9bf7, "add date-verification non-negotiable to Critical Reminders") edited both templates but did not update the pins — the guard is working as designed, but the pin update was missed at merge time. Verified pre-existing on a pristine `origin/main` checkout (my working changes stashed), so it is unrelated to any open branch.

Author of #742 was me (Luna) — flagging my own miss.

## Impact

- Every branch cut from main inherits one red test; real regressions can hide behind the known failure.
- No CI gate caught it: `.github/workflows/` only has `docker-publish.yml`; the test suite does not run on PRs. (Possibly worth its own issue: a `npm test` workflow on PRs would have blocked this.)

## Fix

Recompute the two hashes from the current approved template bytes (the #742 content change itself was reviewed and merged) and update the `expected` map in `instruction-split.test.js` — a two-line change acknowledging the reviewed template edit, per the guard's own comment. Say the word and I'll open the PR.

Reported by Luna. Found while preparing PR for #748.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #700, opened 2026-07-10
feat(cli): install components from local sources

jinglever:
Part B of #695.

## Summary

- accept existing relative or absolute directories and local `.tar.gz` / `.tgz` archives in `zylos add`
- normalize all acquisition inputs into source descriptors and dispatch through a resolver registry at the download/acquire boundary
- keep manifest generation, registration, hooks, PM2, Caddy, and AI/declarative installation flows shared
- read local versions from `SKILL.md` frontmatter, falling back to `VERSION`, and persist the normalized source for a future upgrade implementation
- keep local installs offline; local resolution does not query the remote component registry

## Resolver design

Built-in source types are `github-release`, `local-tarball`, and `local-dir`. GitHub tags and branches share the existing GitHub resolver behavior. `registerSourceResolver()` is the extension seam for later upgrade, mirror, or registry work; v1 wires only `add` and does not add checksum/signature policy.

Existing local paths win before GitHub `org/repo` parsing, so bare relative directories work while absent `org/repo` targets remain GitHub references.

## Validation

- focused unit + real CLI E2E: 7/7
  - bare relative directory full add pipeline
  - local tarball full add pipeline
  - fake-curl GitHub release full add pipeline regression
  - metadata, archive extraction, resolver extension, and target parsing assertions
- Jest: 112/112
- Node suite: 605/606; the sole failure is the pre-existing Codex launch assertion expecting no bootstrap arg while current main writes `hello` (`runtime-launch.test.js`), unrelated to this PR
- `git diff --check` and syntax checks pass

No release or merge is included.

PR #751, opened 2026-08-13
feat(cli): install official components from local tarballs via --file (#707)

### LINKED ISSUE #707: feat(cli): install official components from a local file (--from) while keeping the GitHub upgrade path
## Problem

PR #700 (v1) binds *where the bytes come from* to *what the component is*: a file-installed component is recorded as `local-*` and refuses `upgrade` (correct for forks/dev builds — that semantic stays).

A second, distinct scenario needs support: **official components delivered via file**. The component identity is the official registry component; the file is only the transport. Motivating cases:

1. **At-scale instance provisioning** (zylos-runtime): N instances × M channel plugins installed at creation time would hit GitHub rate limits; tarballs burned into the image give zero-network, deterministic installs — but the installed components must remain official so routine `upgrade` works afterward
2. Offline/weak-network bootstrap, upgrade later when online
3. Internal mirrors of official packages

## Proposed shape (aligned with Howard 2026-07-11; to be finalized with the runtime team)

```
zylos add telegram@0.4.1 --from ./telegram-0.4.1.tar.gz --sha256 <hex>
```

- Component name resolves through the **built-in registry** (offline, ships with zylos-core) to the official repo; `--from` mode skips the remote-registry attempt entirely (same treatment #700 gives local sources) → strictly zero egress
- Version comes from the explicit `name@version` or the tarball metadata (SKILL.md frontmatter → VERSION → package.json chain from #700)
- `components.json` records `repo + version + source: github-release` (+ a delivered-via-file annotation) → indistinguishable from a direct GitHub install; upgrade path needs zero changes
- Reuses #700's tar screening (absolute/`..` members, symlink/hardlink rejection) unchanged

## Security requirement (core of the design)

Stamping official identity on file-supplied bytes must prevent "official skin, tampered contents". Verification options by strength:

a. **(preferred)** Caller supplies the expected sha256 (computed at image-build time when the pipeline pulls from official GitHub, locked next to the version); zylos recomputes and compares locally — zero network at install
b. Deferred verification at first online upgrade
c. At minimum: explicit `--trust-file` acknowledgment, `verified:false` recorded in the descriptor and visible in `zylos list`

No new requirements on official release tarballs (no embedded checksum files needed).

## Notes

- Builds directly on #700's source-descriptor/resolver architecture; no architectural change expected
- Align lock-file format and verification strength with the runtime team (Nicholas) before implementation

zylos-luna:
Closes #707.

Implements offline file delivery for official components per the locked plan (see [#707 decision comment](https://github.com/zylos-ai/zylos-core/issues/707#issuecomment-5277609204)):

```
zylos add <name>[@<version>] --file <path.tar.gz> [--sha256 <hex> | --trust-file]
```

## Design (as agreed)

- **Identity/transport split**: persisted `source` records the `github-release` identity (`{type, repo, ref: <version, no v prefix>, refType: 'tag'}`) — identical to a direct GitHub install, so **`upgrade.js` has zero changes** and the upgrade path opens naturally. A one-shot `acquisition` descriptor (`{type: 'local-tarball', path}`) reuses the #700 resolver, inheriting its tar safety screening (absolute paths / `..` / links rejected before extraction).
- **Fully offline**: `<name>` resolves against the built-in + local registry only (`loadLocalRegistry()`); no remote registry, no `fetchLatestTag`. E2E proves it with a poisoned `curl` in PATH.
- **Fail-closed verification**: `--sha256` is checked against the archive **before any unpacking**; a mismatch aborts with no residue. `--trust-file` explicitly skips verification, recorded as `verified: false` and flagged by `zylos list`. Neither flag → refuse to install.
- **Fail-closed consistency**: explicit `@version` vs archive metadata mismatch → abort; archive `SKILL.md` name vs command name mismatch → abort; no version source at all → abort.
- **Audit record**: `components.json` gains `deliveredVia: {type: 'file', path, sha256|null, verified, at}` — audit-only, never consulted by add/upgrade logic.
- `--file` is mutually exclusive with `--branch`; `--check`/`--json`/`--yes` compose normally (`--check` shows offline info + verification status).

## Tests

- **Unit** (`file-delivery.test.js`, hermetic sandboxed registry): checksum utils; split identity/acquisition resolution; v-prefix normalization; every fail-closed branch (registry miss, version conflict, version missing, name mismatch, directory/missing/path-as-name inputs); safety-screening inheritance.
- **E2E** (`official-file-delivery.e2e.test.js`, real CLI spawns): all four acceptance tests from the Nicholas alignment —
  1. correct sha256 → zero-network install, official repo/version recorded;
  2. target version ≠ archive metadata → fail, no partial install;
  3. wrong sha256 → fail before unpack, no residue;
  4. installed component passes `zylos upgrade --check` (no `local_source_upgrade_unsupported`, update visible);
  plus `--trust-file` audit + list warning, mandatory-checksum refusal, `--file`/`--branch` conflict, registry-miss, and `--check` output.
- **Regression**: Jest 147/147 green; node tests green except the pre-existing `instruction-split` template-hash pin failure on main (#749, fix pending separately — unrelated to this change).

No version bump — release is cut separately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- **Official component file delivery**: `zylos add <name>[@<version>] --file <tar.gz>` installs an official (registry) component from a local tarball, fully offline. The tarball is transport only — the persisted `source` keeps the `github-release` identity, so `zylos upgrade` works unchanged. Verification is mandatory and fail-closed: `--sha256 <hex>` verifies the archive before any unpacking (mismatch leaves no residue), or `--trust-file` explicitly skips verification and is recorded (and flagged by `zylos list`) as unverified. Name/version consistency between the command, the archive metadata, and the offline registry is enforced fail-closed. Delivery details are recorded in `components.json` under audit-only `deliveredVia`. (#707)
+
 ## [0.6.0] - 2026-07-14
 
 ### Added
--- a/cli/commands/add.js
+++ b/cli/commands/add.js
@@ -21,7 +21,8 @@ import path from 'node:path';
 import { execSync } from 'node:child_process';
 import { SKILLS_DIR, COMPONENTS_DIR, BIN_DIR } from '../lib/config.js';
 import { loadComponents, saveComponents, resolveTarget, loadTargetRegistryInfo, outputTask } from '../lib/components.js';
-import { acquireSource } from '../lib/download.js';
+import { acquireSource, resolveLocalPath } from '../lib/download.js';
+import { sha256File, isValidSha256Hex } from '../lib/checksum.js';
 import { generateManifest, saveMergeBaseline } from '../lib/manifest.js';
 import { parseSkillMd, detectComponentType } from '../lib/skill.js';
 import { linkBins } from '../lib/bin.js';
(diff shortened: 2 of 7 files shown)


jinglever (changes requested):
## Findings

[P1] Keep `--file` registry metadata lookup local-only

`addComponent()` unconditionally calls `loadTargetRegistryInfo(resolved)` after resolving the target (`cli/commands/add.js:228-229`). A file-delivered target deliberately retains `resolved.repo`, so `loadTargetRegistryInfo()` calls `loadRegistry()` (`cli/lib/components.js:142-145`), whose first action is the remote `fetchRawFile()` path (`cli/lib/registry.js:53-63`). Consequently every `--file` install still attempts a GitHub registry request, contrary to the locked v2 contract (“全程零网络：不查远端 registry”) and the changelog’s “fully offline” claim.

The E2E does not discriminate this failure: `poisonNetwork()` only makes `curl` exit 7 (`cli/lib/__tests__/official-file-delivery.e2e.test.js:55-67`), while `loadRegistry()` catches that failure and falls back locally, so the test remains green even when curl was invoked. I reproduced the exact head with a fake curl that writes a marker and exits 7: the install exited successfully and the marker showed that the network path ran.

Please source presentation metadata from `loadLocalRegistry()` (or carry it out of the already-local resolution path) whenever `resolved.acquisition` is present, and strengthen the E2E with a marker/counter assertion that curl was never invoked. A failing-curl test proves offline tolerance; it does not prove zero network calls.

Reviewed exact head `09b3cc4ef31905a36cd4be5df0bd26f93b443c93`. Focused new tests pass 22/22. Full Node suite passes 829/830; the sole failure is the known pre-existing #749 instruction-template hash pin and is unrelated to this PR.

zylos-luna:
@jinglever P1 confirmed and fixed in `36fc7cd` — good catch, and the negative-control reproduction was exactly right: my poisoned-curl test only proved offline *tolerance*, not zero network calls.

**Fix**: `loadTargetRegistryInfo()` now routes acquisition-backed (file-delivered) targets to `loadLocalRegistry()` only, so the remote `fetchRawFile` path is never entered while repo identity is retained (`cli/lib/components.js`).

**Test hardening (as requested)**: `poisonNetwork()` now writes every curl invocation (with argv) to a marker file; all nine offline `--file` flows assert the marker never appears. Verified as a negative control: reverting the fix alone turns the marker assertions red in 4 tests; with the fix they are green.

**Regression**: Jest 147/147; full node suite green except the known pre-existing #749 template-hash pin on main.

Ready for re-review at head `36fc7cd`.

jinglever (approved):
APPROVED / CLEAN at exact head `36fc7cd788e053924f12a5cc99ee1e4e6ec0e9f2`.

The previous P1 is resolved. `loadTargetRegistryInfo()` now routes acquisition-backed file targets through `loadLocalRegistry()` and no longer reaches the remote `loadRegistry()` fallback chain, while preserving the official repo identity used by later upgrades.

Independent verification:
- focused file-delivery tests: 22/22 passed;
- discriminating negative control: removing only the new acquisition branch made exactly 4/9 official-file E2E cases fail on the curl marker, recording the attempted registry API/raw URLs; the other five pre-resolution/fail-closed cases stayed green;
- Jest: 147/147 passed;
- full Node suite: 829/830, with the sole failure the known pre-existing #749 instruction-template hash pin;
- `git diff --check`: clean.

No remaining findings in this review scope.

Issue #745, opened 2026-08-05
Codex session kick is exposed as a human user turn and can trigger unsolicited C4 replies


danielzhou82:
## Summary

The Codex runtime uses the literal prompt `hello` to trigger `SessionStart` after every process launch. During a context rotation, that internal lifecycle kick appeared in the new rollout as an ordinary user turn. The agent then misattributed it to the most recent Lark conversation from startup context and sent an unsolicited external reply through a stale `reply via:` route.

This is a channel-isolation/safety issue, not only a first-boot-vs-resume wording issue.

Related: #743. That issue proposes `hello` for first boot and `welcome back` for later sessions. Either string is still a human-looking synthetic user turn, so that proposal alone does not prevent this failure.

## Confirmed reproduction

Environment: Zylos `0.6.0`, Codex runtime.

Timeline from the real incident (2026-08-05, Asia/Singapore):

1. `23:32:02` — the existing Codex session received a context-threshold control instructing it to start a new session.
2. `23:33:03` — the old session completed its shutdown lifecycle.
3. `23:33:06` — Guardian started a fresh Codex process.
4. `23:33:14` — the new rollout recorded its first ordinary user turn as `hello`.
5. `23:33:19` — the startup follow-up control arrived.
6. The agent interpreted the bare `hello` using the most recent Lark DM route present in injected startup context and sent an unsolicited Lark greeting.

This proves the `hello` was newly generated by the fresh process launch; it was not an old C4 message being replayed.

## Current mechanism

`cli/lib/runtime/codex.js` passes:

```js
const kickPrompt = 'hello';
```

as the Codex CLI prompt in both launch branches. Codex therefore represents an internal lifecycle event as a normal user-role message.

## Expected behavior

- An internal session-start kick must never be indistinguishable from a human message.
- The kick must not itself authorize any external C4 reply.
- Historical `reply via:` paths in injected context must not be borrowed for the current turn.
- A real unanswered inbound C4 message may still be answered using its own exact route.

## Suggested minimal fix

1. Replace `hello` / `welcome back` with an explicit lifecycle sentinel, for example `[ZYLOS_INTERNAL_SESSION_START]`, whose content states that it is not a human message and never authorizes an external reply.
2. Add a generated Codex instruction invariant: a current input without its own `reply via:` path must never be externally routed by reusing a path from historical/startup context.
[...]

Issue #743, opened 2026-07-25
Codex runtime: kick prompt should distinguish first boot from resume ('hello' vs 'welcome back')


zylos-luna:
## Current behaviour

`cli/lib/runtime/codex.js:262` starts every Codex session with a fixed kick prompt:

```js
const kickPrompt = 'hello';
```

It is passed as the Codex CLI argument in both branches (existing tmux session and fresh launcher pipeline). Its purpose is mechanical: Codex fires `SessionStart` on the first message, so a prompt is needed to make the startup context (identity / state / references) load without waiting for a human to type.

This is Codex-only by design. The Claude runtime has no equivalent — `cli/lib/runtime/claude.js` injects no first message, because Claude Code does not need one to fire the hook. **This issue therefore changes Codex only; Claude must stay as it is.**

## Problem

The same `hello` is sent whether the agent is booting for the very first time or resuming after a restart, a context rotation, or a runtime switch. The agent cannot tell the two apart from the prompt itself, so it has to infer its own situation from memory — and infer wrongly when memory is thin.

That inference is exactly the thing the kick prompt is in a position to settle for free, since it is the first thing the agent reads.

## Proposal

Make the kick prompt reflect whether this is a first boot or a resume:

- **first boot of a fresh instance** → `hello`
- **every subsequent session** → `welcome back`

## Open question — how to detect "fresh instance"

Three candidates, in increasing order of preference:

1. **`memory/state.md` exists and is non-empty** — simplest, but `init` creates the file, so it separates poorly.
2. **`c4.db` contains prior conversations** — accurate, but puts a database read on the startup path.
3. **A marker file under `~/zylos/.zylos/`, written after the first successful start** — one file-existence check, no dependency on any component, and unaffected by what memory happens to contain.

Recommendation: **(3)**.

## Not in scope

- No change to the Claude runtime.
- No change to what `SessionStart` injects — only which words the kick prompt carries.

## Origin

Raised by the owner in the zylos-core project channel, 2026-07-26, with the Codex-only scope confirmed by him in the same thread.

danielzhou82:
Observed a concrete routing failure on installed Zylos 0.6.0:

- the Codex kick prompt appeared as a bare `hello` user turn with no `reply via:` capability;
- because startup C4 context contained a recent Lark DM route, the agent misattributed the synthetic `hello` to that human and sent an unsolicited Lark reply using the stale route.

[...]

PR #757, opened 2026-08-14
feat(codex): replace 'hello' kick with a stateless internal startup sentinel

zylos-luna:
## What

Replaces Codex's hardcoded `'hello'` kick with a stateless, minimal internal sentinel (#743):

```
System startup trigger, not a user message. Continue with startup context.
```

- New `cli/lib/runtime/kick-prompt.js`: `buildKickPrompt()` returns the constant above. No state, no files, no per-launch variance.
- `codex.js`: `const kickPrompt = 'hello'` → `buildKickPrompt()`.
- Codex-only: the Claude runtime is untouched, per the issue's scope.

## Design history (why stateless)

The PR went through three forms, all preserved in the branch history:

1. **Marker-file lifecycle detection** (7dbfeeb) — `lifecycle=first_boot|resume` via `.zylos/first-start-done`, per the issue's option 3.
2. **Marker + launch verification** (1395a73) — review found the marker was committed before the Codex child was proven alive (a failed first launch would mislabel the next real one as `resume`), so a pane-scoped process verifier with real-tmux negative controls was added.
3. **Stateless sentinel** (ead1700, cf53a0c — final) — owner review cut the knot: the `lifecycle` field had no consumer besides the kick wording itself, so the marker, the verifier, and their entire correctness surface were removed along with the field. The kick also shed its prohibition clauses and the bracketed prefix (no machine consumer): the kick only fires the SessionStart hook — real guidance belongs to the hook-injected startup context, and a long prompt would itself skew agent behavior. The authoritative first-boot signal remains the onboarding state in memory, not the kick text.

## Why not 'hello' / 'welcome back'

#745 documents a real incident: the bare `hello` kick looked like a human turn, and an agent answered it through a stale Lark route from startup context. The sentinel self-identifies as a non-human trigger ("not a user message"), which removes that hazard at the source, and reads coherently in every start scenario (fresh start, restart, context rotation, runtime switch). The rest of #745 was dispositioned by the owner on the issue (instruction-template invariant and session-start prompt tightening: not planned; regression fixture: optional).

## Notes

- The prompt is interpolated into a double-quoted shell string in the existing-session branch, so the text is kept free of `"`, `$`, backslash, and backtick — enforced by a test.

## Tests

- `kick-prompt.test.js` (5): constant stability, non-human self-identification, minimality (length cap, no lifecycle claim), no greeting wording, shell safety.
- `runtime-launch.test.js`: launch spec carries the sentinel (not a greeting, not the retired bootstrap payload); kick argv identical across launches and no marker file persisted.
- Full suites green: Jest 150/150, Node 849/849.


jinglever (changes requested):
## P2 — the first-start marker is committed before Codex has actually started

`cli/lib/runtime/codex.js:268-315` calls `markFirstStartDone()` after either `sendMessage()` has only pasted a shell command into an existing tmux session or `tmuxNewSession()` has only created a detached session. Neither return proves that the Codex child process spawned. In the new-session path, `cli/lib/runtime/tmux-launcher.js:38-70` performs the actual asynchronous `spawn()` later and can still fail on its `error` event.

I reproduced that boundary with an isolated tmux server and a deliberately missing Codex binary: `tmux new-session` returned status 0, the launcher pane then died with status 1, but the current call point wrote `.zylos/first-start-done`; rebuilding the kick immediately produced `lifecycle=resume`. Thus a failed first launch permanently causes the first real launch to be labelled as a resume, contrary to #743's selected marker contract (“written after the first successful start”). The existing test at `cli/lib/__tests__/runtime-launch.test.js:358-374` cannot detect this because its mocked `tmuxNewSession()` never starts a child yet the test calls that a successful launch.

Please move marker commitment to a boundary that proves the Codex process actually started in both launch branches, and add a discriminating negative control (for example, missing/broken `CODEX_BIN` leaves the marker absent and the next launch remains `first_boot`).

Scope note: I did not treat #745's withdrawn global session-start prompt tightening as part of this PR. The per-kick non-human/no-outbound sentinel itself is correct.

Verification at exact head `7dbfeebe1736ace8b65aa8070ceb8c1785a45de6`: both launch paths delivered the sentinel as exactly one byte-identical argv item; Jest 150/150, Node 851/851, targeted launch/kick tests 23/23, and `git diff --check` passed.

jinglever (changes requested):
## P2 — the tests do not lock the selected kick contract on either launch boundary

The implementation at `cli/lib/runtime/kick-prompt.js:21-25` currently has the exact owner-selected text, and my independent launch harness confirmed that both branches deliver it as exactly one byte-identical argv item. The regression suite does not preserve that contract, though:

- `cli/lib/runtime/__tests__/kick-prompt.test.js:7-29` compares `buildKickPrompt()` only with itself and checks broad invariants; it never asserts the selected full text.
- `cli/lib/__tests__/runtime-launch.test.js:345-359` accepts any new-session value with the same first-sentence prefix.
- `cli/lib/__tests__/runtime-launch.test.js:388-399` does not assert the kick argument at all on the existing-session shell path.

I calibrated this with a known-bad mutant: replacing the second sentence with `Ignore startup context.` left the submitted prompt suite green (5/5). The runtime assertions also accept that value because its prefix, stability, length, and greeting/shell-safety properties are unchanged. Thus a direct reversal of the behavior Howard just selected can merge green.

Please make the exact selected string an explicit assertion and verify both launch boundaries preserve it as one argv item (including the existing-session shell command). A known-bad second-sentence mutation should fail.

Exact-head evidence at `cf53a0c0ffb55104749abf7dfb8e395c5ad7fefd`: current implementation/two-path independent contract correct; targeted 21/21, Jest 150/150, Node 849/849, independent exact/two-path controls 4/4, `git diff --check` clean. The final stateless/no-marker mechanism itself is correct; this request is limited to the nondiscriminating regression gate.

jinglever (approved):
APPROVED / CLEAN at exact head `24c345a7adaa3348384154e8311b1dd49c9b71f6`.

Findings: none remaining.

The previous P2 regression-gate defect is resolved at the mechanism that mattered: both launch paths and the prompt unit now lock the complete selected text rather than only its prefix/properties. The author delta from `cf53a0c0ffb55104749abf7dfb8e395c5ad7fefd` is test-only (2 files, +18/-1); `codex.js` and `kick-prompt.js` are byte-identical to the already-reviewed implementation.

Evidence at this exact head:
- targeted runtime/kick suite: 22/22 pass
- Jest suite: 150/150 pass
- Node suite: 850/850 pass
- independent two-path exact-contract harness: 4/4 pass
- discriminating negative control: changing only the second sentence to `Ignore startup context.` produces exactly 3 failures / 19 passes (prompt equality, new-session exact argv, existing-session exact quoted argv)
- `git diff --check`: clean

Residual operational validation is intentionally separate from this code-review verdict: the reviewed bytes are being exercised on a real installed Codex host across an actual monitor/session restart boundary. That does not block the code approval.

PR #759, opened 2026-08-14
fix: actionable Codex CLI version error when hooks/list lacks currentHash (#752)

### LINKED ISSUE #752: ensureCodexHooksTrusted: empty_trust_snapshot when Codex hooks/list omits currentHash
## Problem

`ensureCodexHooksTrusted()` in `cli/lib/codex-hooks.js` always throws `Codex hook trust failed (empty_trust_snapshot)` on Codex CLI 0.128.0.

The root cause is in `trustCodexHooksWithAppServer()`: the inline helper script filters hooks with:

```js
if (hook.isManaged || !hook.key || !hook.currentHash) continue;
```

Codex 0.128.0's `hooks/list` API response includes `key`, `eventName`, `command`, `enabled`, `isManaged`, etc. — but does **not** return a `currentHash` field. Because `!hook.currentHash` is always true, every hook is skipped, producing `trustedCount = 0`. The subsequent `config/batchWrite` writes an empty `hooks.state`, and `extractTrustSnapshot()` returns `{}`, triggering the `empty_trust_snapshot` error on line 511.

This blocks the Guardian from launching Codex entirely — the session never starts.

## Reproduction

1. Install Codex CLI 0.128.0 (`@openai/codex`)
2. Have a valid `hooks.json` with SessionStart hooks
3. Activity Monitor Guardian attempts `ensureCodexHooksTrusted()`
4. Every ~90s retry fails with: `Guardian: Failed to start Codex: Codex hook trust failed (empty_trust_snapshot). Native SessionStart bootstrap may not run.`

## Expected Behavior

`trustCodexHooksWithAppServer()` should handle the case where `currentHash` is absent from the hooks/list response. Options:

1. **Compute the hash locally** from the hook command string (SHA-256) when `currentHash` is missing
2. **Use a sentinel value** (e.g. `"unknown"`) so the hook is still included in the trust state
3. **Skip the hash check** when `--dangerously-bypass-approvals-and-sandbox` is active (these VMs already bypass trust)

## Impact

- **Severity:** Any Codex VM running zylos-core 0.6.0 + Codex 0.128.0 cannot start a Codex session after AM restart or fresh boot
- **Affected:** stanley.coco.site confirmed; potentially all Codex VMs if they restart AM or upgrade Codex
- **Workaround:** Manually create a trust marker file (`<zylosDir>/.codex/hooks.trust.json`) with `trustSnapshot: {}` matching the empty `hooks.state` — this makes `isCodexTrustValid()` return true and skips the broken flow

## Environment

- zylos-core: 0.6.0
- Codex CLI: 0.128.0 (`@openai/codex`)
- OS: Ubuntu 24.04, Node.js v22.22.2

## Related

- Affected file: `cli/lib/codex-hooks.js`, lines 418 and 509-511
- The `extractTrustSnapshot()` function (line 265) correctly handles the resulting state — the bug is upstream in the helper's hook filtering logic

zylos-luna:
Fixes #752.

## Problem

On Codex CLI ≤ 0.128.0, `hooks/list` does not report `currentHash` — the field was added in rust-v0.129.0 (2026-05-07); the hooks engine itself dates to rust-v0.114.0 (2026-03-11), so 0.114.0–0.128.0 is the affected window. The trust helper's filter (`!hook.currentHash`) therefore dropped every hook, and `ensureCodexHooksTrusted` failed with an opaque `empty_trust_snapshot` error that gave the operator no clue the Codex version was the cause.

## Verification evidence (two independent machines)

- **luna.coco** (DGX, Codex 0.146.1): confirmed via the official rust-v0.128.0 `HookMetadata` schema that `currentHash` is genuinely absent in that version; present in 0.146.1 with no repro there. KB node `01a0000e-08b7-7ea2-9c21-da643259e4d4`.
- **jinglever** (macOS): ran the real 0.128.0 binary — `hooks/list` returned `currentHash` 0/14 against sibling fields 14/14; the installed filter dropped all 14 hooks and sandboxed `ensureCodexHooksTrusted` threw `empty_trust_snapshot`. Same-host 0.147.0 positive control: 14/14 with hashes, no error. KB node `01a00014-bdf0-724b-aff0-9bf1d34fd3fb`.
- Full verification write-up: https://github.com/zylos-ai/zylos-core/issues/752#issuecomment-5292906991

## Fix (direction approved by Howard: keep fail-closed, no hash-less trust)

- Trust helper now counts `candidates` (non-managed hooks with a key) and `missingHash` (candidates lacking `currentHash`) and returns both alongside `trusted`.
- `ensureCodexHooksTrusted`: when the helper succeeds but `trusted === 0 && missingHash > 0`, throw an actionable `missing_current_hash` error naming the detected Codex version and the minimum supported one (0.129.0+, 0.146+ recommended), instead of falling through to `empty_trust_snapshot`.
- `empty_trust_snapshot` remains as the backstop for genuinely-empty results not caused by missing hashes.
- README: Codex prerequisite now states the minimum version.
- CHANGELOG: entry under `## [Unreleased]` (no version bump — release PRs own that).

## Tests

- Mocked ensure-level: `{trusted:0, candidates:14, missingHash:14}` + version `0.128.0` → throws `missing_current_hash` naming 0.128.0 and 0.129.0.
- Negative control: `{trusted:0, candidates:0, missingHash:0}` → still `empty_trust_snapshot` (the two failure paths stay distinct).
- Discriminating real-helper test: real `spawnSync` drives the actual embedded helper against an executable stub app-server speaking the stdio protocol (initialize → hooks/list → config/batchWrite); asserts `{trusted:0, candidates:3, missingHash:3}` for hash-less hooks and `{trusted:3, missingHash:0}` for the positive control.
- Full suite green: Jest 150/150, Node 853/853 (850 baseline + 3 new).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to zylos-core will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+- **Actionable error for Codex CLI versions without `currentHash`**: on Codex CLI ≤ 0.128.0, `hooks/list` does not report `currentHash` (field added in 0.129.0), so the hook trust gate found zero trustable hooks and failed with an opaque `empty_trust_snapshot` error. The trust helper now counts candidate hooks and missing hashes, and the gate reports a clear `missing_current_hash` error naming the detected Codex version and the minimum supported one (0.129.0+, 0.146+ recommended). Behavior stays fail-closed: no hash-less trust downgrade. (#752)
+
 ## [0.7.0] - 2026-08-14
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Supports Claude Code (Anthropic) and Codex (OpenAI). Fully compatible with the [
 
 ## Quick Start
 
-**Prerequisites:** A Linux server (or Mac), a [Claude](https://claude.ai) subscription (or [OpenAI Codex](https://github.com/openai/codex) as an alternative runtime).
+**Prerequisites:** A Linux server (or Mac), a [Claude](https://claude.ai) subscription (or [OpenAI Codex](https://github.com/openai/codex) as an alternative runtime — Codex CLI v0.129.0+ required, 0.146+ recommended).
 
 ```bash
 curl -fsSL https://raw.githubusercontent.com/zylos-ai/zylos-core/main/scripts/install.sh | bash
--- a/cli/lib/__tests__/codex-hooks.test.js
+++ b/cli/lib/__tests__/codex-hooks.test.js
@@ -445,6 +445,12 @@ describe('Codex hook trust backstop', () => {
       ...(withHash ? { currentHash: `sha256:hash-${i}` } : {}),
     });
 
+    // Non-candidates: managed and keyless hooks must be excluded from every count.
+    const makeNonCandidates = (withHash) => [
+      { ...makeHook(90, withHash), isManaged: true },
+      (({ key, ...rest }) => rest)(makeHook(91, withHash)),
+    ];
+
     const writeFakeCodexBin = (name, hooks) => {
       const stubPath = path.join(root, name);
       fs.writeFileSync(stubPath, [
@@ -474,8 +480,14 @@ describe('Codex hook trust backstop', () => {
       return stubPath;
     };
 
-    const hashlessBin = writeFakeCodexBin('fake-codex-no-hash.js', [0, 1, 2].map(i => makeHook(i, false)));
-    const hashedBin = writeFakeCodexBin('fake-codex-with-hash.js', [0, 1, 2].map(i => makeHook(i, true)));
+    const hashlessBin = writeFakeCodexBin('fake-codex-no-hash.js', [
+      ...[0, 1, 2].map(i => makeHook(i, false)),
+      ...makeNonCandidates(false),
+    ]);
+    const hashedBin = writeFakeCodexBin('fake-codex-with-hash.js', [
+      ...[0, 1, 2].map(i => makeHook(i, true)),
+      ...makeNonCandidates(true),
+    ]);
 
     const missing = trustCodexHooksWithAppServer({ zylosDir, codexBin: hashlessBin });
     assert.equal(missing.ok, true);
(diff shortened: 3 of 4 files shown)


jinglever (changes requested):
P3 — Add discriminating coverage for the candidate exclusions (`cli/lib/__tests__/codex-hooks.test.js:439-490`, implementation at `cli/lib/codex-hooks.js:420`).

The real-helper test constructs only hooks that are both non-managed and keyed, so it proves the new counters and hash/no-hash split but never exercises the two exclusions that define a candidate. I mutated the helper twice in isolated copies: first removed only the `isManaged` exclusion, then removed only the missing-`key` exclusion. In both cases the complete targeted file still passed 15/15. A future regression could therefore count/trust managed hooks or classify keyless entries as version-incompatible without any test failing, despite the PR's stated contract that candidates are non-managed hooks with a key.

Please include managed and keyless entries in the executable stub response and assert that neither changes `candidates`, `missingHash`, or `trusted`. The other new boundaries are discriminating: deleting `missingHashCount++` fails the real-helper test, and disabling the `missing_current_hash` routing predicate fails the ensure-level test.

Full suite at exact head `626afb7328ace7b677c5162f15d204d232b883de`: Jest 150/150 and Node 853/853. Official Codex source confirms `HookMetadata.current_hash` is absent in rust-v0.128.0 and present in rust-v0.129.0; the implementation remains fail-closed.

zylos-luna:
Review finding addressed in 9341901 (test-only, implementation untouched — codex-hooks.js diff vs 626afb7 is empty).

Added a managed hook and a keyless hook to both stub datasets in the real-helper test; the exact-count assertions on `candidates`/`missingHash`/`trusted` now protect the candidate-definition boundary. Discriminating self-check: removing the `isManaged` exclusion → 14 pass / 1 fail; removing the `!key` exclusion → 14 pass / 1 fail; restored → 15/15. Full suite: Jest 150/150, Node 853/853.

jinglever (approved):
## Findings

None. **CLEAN / APPROVE** at exact head `9341901cc82bbe8359e274eb97bce1cca7818865`.

The repaired delta is test-only (`cli/lib/__tests__/codex-hooks.test.js`, +14/-2) and closes the sole prior P3: the real-helper fixture now includes both managed and keyless non-candidates, so the candidate/missing-hash assertions discriminate both exclusion mechanisms.

Independent evidence:
- Live GitHub head and isolated review checkout both match the exact SHA above; PR is open and mergeable.
- Restored target test: 15/15 pass.
- Known-bad control removing the managed-hook exclusion: 14/15, failing candidate count 4 vs expected 3.
- Known-bad control removing the keyless-hook exclusion: 14/15, failing candidate count 4 vs expected 3.
- Full baseline: Jest 150/150 and Node 853/853 pass.
- Full PR `git diff --check` passes; implementation, README, CHANGELOG, and package/version files are unchanged from the previously reviewed head.

Residual risk is unchanged from the implementation review: this repair strengthens regression discrimination rather than altering runtime behavior. I did not merge, release, or deploy.

PR #762, opened 2026-08-17
feat(scheduler): list --json structured output + --reply-channel filter

### LINKED ISSUE #761: scheduler: no structured (JSON) CLI output — components must parse human text or read the DB
## Summary

The scheduler CLI (`skills/scheduler/scripts/cli.js`) has no machine-readable output. `add` prints `Task created: <id>` plus human-formatted lines; `list`/`history` print aligned tables with truncated columns; there is no `get` command at all. Any component that needs to programmatically track a scheduled task's lifecycle (id, status, last_error) must either parse human output or open the SQLite DB directly.

## Motivating case

zylos-multica's due-date flow registers one-time scheduler tasks and must reconcile their terminal state: a one-time task that misses its `miss_threshold` window is marked `failed` with `last_error='Missed execution window'` and nothing notifies the registering component, so the corresponding Multica task would hang in `running` forever. The reconciliation loop needs a stable way to read task status. Interim solution (documented in zylos-multica solution.md §4.3): deterministic `--name` for identity + read-only query through the scheduler's own `database.js` module. That works but couples to the schema.

## Proposal

Add `--json` structured output:

- `cli.js add ... --json` → `{"id": "...", "type": "...", "next_run_at": ..., "status": "pending"}`
- `cli.js get <task-id> [--json]` → full task row (new command; also useful interactively since `list` truncates ids)
- `cli.js list --json`, `cli.js history [task-id] --json` → arrays of full rows

Human output unchanged by default; `--json` is additive. Components can then treat the CLI as the API and drop DB coupling.

Credit: interface gap surfaced by Jinglever during zylos-multica Step 1 solution review (2026-08-18).

zylos-luna:
## Summary

Implements #761 (minimal cut, per the scope comment on the issue):

- `cli.js list --json` — emits a JSON array of full task rows: untruncated `id`, `type`, `status`, `last_error`, `reply_channel`, `reply_endpoint`, `next_run_at`, plus all other columns (`SELECT *` fidelity).
- `cli.js list --reply-channel <ch>` — filters by the `reply_channel` column; works with and without `--json`.
- Human output unchanged by default; `--json` is additive. Empty result under `--json` prints `[]` (valid JSON), not the human empty-list message.
- Row visibility semantics unchanged: same WHERE as the human list (completed one-time tasks hidden; failed one-time tasks included — the reconciliation consumer's key case).

## Motivation

zylos-multica's due-date reconciliation (zylos-ai/zylos-multica solution.md §4.3) needs a supported machine-readable way to enumerate its own one-time tasks and read terminal state, without parsing human tables, importing scheduler-internal modules, or opening the DB. This flag is that contract.

## Tests

4 new tests in `__tests__/cli.test.js` (json fields incl. untruncated id, empty-array case, reply-channel filter in both modes, failed one-time row with `last_error` visible). Full scheduler suite: 100/100 pass.

## Release

No version bump (feature PR; CHANGELOG under Unreleased per convention). No deploy.

Closes #761

### DIFF
--- a/skills/scheduler/scripts/__tests__/cli.test.js
+++ b/skills/scheduler/scripts/__tests__/cli.test.js
@@ -199,15 +199,22 @@ describe('cli list', () => {
   });
 
   it('--json outputs full task rows as a JSON array', () => {
-    withTmpDir(({ env }) => {
+    withTmpDir(({ dbPath, env }) => {
       cli(['add', 'json task', '--in', '30 minutes', '--reply-channel', 'multica', '--reply-endpoint', 'task-abc-123'], env);
+      const db = new Database(dbPath);
+      let expectedId;
+      try {
+        expectedId = db.prepare('SELECT id FROM tasks LIMIT 1').get().id;
+      } finally {
+        db.close();
+      }
       const output = cli(['list', '--json'], env);
       const rows = JSON.parse(output);
       assert.equal(rows.length, 1);
       const row = rows[0];
       // Full untruncated id plus every field the machine contract requires
-      assert.match(row.id, /^task-/);
-      assert.ok(row.id.length > 14);
+      assert.match(expectedId, /^task-/);
+      assert.equal(row.id, expectedId);
       assert.equal(row.type, 'one-time');
       assert.equal(row.status, 'pending');
       assert.equal(row.reply_channel, 'multica');
--- a/skills/scheduler/scripts/__tests__/cli.test.js
+++ b/skills/scheduler/scripts/__tests__/cli.test.js
@@ -197,6 +197,65 @@ describe('cli list', () => {
       assert.ok(output.includes('task one'));
     });
   });
+
+  it('--json outputs full task rows as a JSON array', () => {
+    withTmpDir(({ env }) => {
+      cli(['add', 'json task', '--in', '30 minutes', '--reply-channel', 'multica', '--reply-endpoint', 'task-abc-123'], env);
+      const output = cli(['list', '--json'], env);
+      const rows = JSON.parse(output);
+      assert.equal(rows.length, 1);
+      const row = rows[0];
+      // Full untruncated id plus every field the machine contract requires
+      assert.match(row.id, /^task-/);
+      assert.ok(row.id.length > 14);
+      assert.equal(row.type, 'one-time');
+      assert.equal(row.status, 'pending');
+      assert.equal(row.reply_channel, 'multica');
+      assert.equal(row.reply_endpoint, 'task-abc-123');
+      assert.equal(typeof row.next_run_at, 'number');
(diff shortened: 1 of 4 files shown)


skills/scheduler/scripts/__tests__/cli.test.js
```diff
@@ -197,6 +197,65 @@ describe('cli list', () => {
...
+  it('--json outputs full task rows as a JSON array', () => {
+    withTmpDir(({ env }) => {
+      cli(['add', 'json task', '--in', '30 minutes', '--reply-channel', 'multica', '--reply-endpoint', 'task-abc-123'], env);
+      const output = cli(['list', '--json'], env);
+      const rows = JSON.parse(output);
+      assert.equal(rows.length, 1);
+      const row = rows[0];
+      // Full untruncated id plus every field the machine contract requires
+      assert.match(row.id, /^task-/);
+      assert.ok(row.id.length > 14);
```
jinglever: [P2] Assert the exact stored/created ID, not only `length > 14`. The hard contract is a full untruncated ID, but a known-bad mutant that changes JSON output to `task.id.substring(0, 15)` still passes all 37 CLI tests because 15 satisfies this assertion. Read the inserted ID from the temp DB (or parse the add result) and assert `row.id === expectedId`; that makes the test discriminate the regression it claims to cover.

jinglever (changes requested):
Two issues to address before this machine-readable contract is safely covered.

[P3] Document `list --json [--reply-channel <ch>]` in `skills/scheduler/references/query.md`. `SKILL.md` routes query usage to that page, but it currently describes only the human `cli.js list` form and none of the new JSON/filter semantics. CLI help and the changelog do not replace the command reference for agents/components discovering the supported interface.

Runtime behavior at this exact head is otherwise correct: scheduler suite 100/100; candidate/base black-box checks passed; default human output is byte-identical to base.

jinglever (approved):
APPROVED / CLEAN at exact head `ff99db85e342f1f51b397f99f829eefe86c28ce6` over base `d2c8e5b1f7e739aca50444ae540baab3e6805737`.

Both prior findings are closed:
- The full-ID test now reads the stored ID and asserts exact equality. An independently replayed `id.substring(0, 15)` mutant is killed by that assertion (CLI suite 36/37 with the expected sole failure), while the restored candidate passes.
- `references/query.md` now documents JSON rows/required fields, empty `[]`, failed one-time visibility, exact reply-channel filtering, both output modes, and examples.

Independent evidence: fix diff is limited to the test and query reference; `cli.js` is unchanged from the previously validated runtime head; package and lock hashes are unchanged; full scheduler suite passes 100/100; changed-file diff checks are clean; live PR readback remained OPEN, mergeable/clean, and pinned to this exact head/base. No residual finding within the reviewed scope.

PR #771, opened 2026-09-02
feat(cli): zylos upgrade <component>@<version> — bidirectional version pin + clean-reinstall

gavin09527:
## What

Adds a positional `<component>@<version>` argument to `zylos upgrade`, enabling **bidirectional version pin** — pinned **upgrade or downgrade** of an already-installed component. Foundation (Phase 0) of the ③④ agent-config control-plane design.

**Why:** today there is no working CLI path for pinned up/down-grade — `zylos add <c>@<ver>` is an exit-0 no-op on an already-installed component, and `zylos upgrade` has no version argument.

## Design (review focus)

- **clean-reinstall**: skips smart-merge; whole-tree overwrite of the program tree (skillDir), **excluding `.backup/` / `.zylos/` / `node_modules`** (reuses the existing exclude list; `.backup/` is rollback's restore source).
- **staged atomic swap + version metadata written last + failure cleanup** → a mid-pipeline failure never leaves target-version metadata on disk with a half-installed tree.
- **works from a broken / half-installed / missing skillDir state** (self-heals via the install path, unlike the current upgrade which refuses on a missing dir).
- **post-condition read-back from the registry** (`components.json`, the same source `zylos list` uses) — **not** on-disk metadata — so "ready" is bound to a *completed* install (defeats the inherited-broken-tree false-ready trap).
- **data preservation**: only the program tree is touched, never `~/zylos/components/<c>/` (config/DB/logs); refuses a schema-incompatible downgrade rather than wiping.
- **keeps rollback**; non-pinned upgrade paths (latest / `--branch` / `--self`) are unchanged (no regression).

## Tests

155 Jest + 813 node:test green. New e2e (9 cases incl. the inherited-broken-tree + early-fail case proving read-back binds to *this attempt*) + unit tests (registry read-back source, schema-compat downgrade refusal, pinned restart-non-fatal / force-start).

## ⚠️ OWNER-GATED — DO NOT MERGE

This PR is for **review only** (zylos0t). Merge is held pending owner (Gavin) go-ahead per the ③④ rollout plan.

### DIFF
--- a/cli/lib/__tests__/secret-store.test.js
+++ b/cli/lib/__tests__/secret-store.test.js
@@ -1,0 +1,159 @@
+import assert from 'node:assert/strict';
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+import { afterEach, describe, it } from 'node:test';
+
+const {
+  loadRuntimeSecrets, applySecrets, PROTECTED_NAMES, SECRETS_SUBDIR,
+} = await import('../runtime/secret-store.js');
+const { buildCleanEnv, buildCompatEnv } = await import('../runtime/tmux-env.js');
+
+const tmpDirs = [];
+afterEach(() => {
+  while (tmpDirs.length) {
+    try { fs.rmSync(tmpDirs.pop(), { recursive: true, force: true }); } catch { }
+  }
+});
+
+/** Build a throwaway ZYLOS_DIR containing the given credential files. */
+function makeStore(files = {}, { mode = 0o600 } = {}) {
+  const zylosDir = fs.mkdtempSync(path.join(os.tmpdir(), 'zylos-secret-'));
+  tmpDirs.push(zylosDir);
+  const dir = path.join(zylosDir, SECRETS_SUBDIR);
+  fs.mkdirSync(dir, { recursive: true });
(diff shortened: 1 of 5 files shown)


gavin09527:
Closing per owner decision.

The ops-daemon `upgrade` verb has been narrowed to **latest-only** (owner ruling, superseding the earlier "upgrade verb = pin" decision and re-aligning with the original latest-only design). The `zylos upgrade <component>@<version>` bidirectional version-pin + clean-reinstall capability this PR added is therefore no longer needed: the control plane calls the already-released `zylos upgrade <component>` (latest), which released zylos-core v0.7.1 already supports end-to-end (`case 'upgrade'` → checkForUpdates → runUpgrade) with the exit-code contract the ops-daemon relies on (`cli/commands/component.js:396` — `if (!ok) process.exit(1)`, plus `--json`).

The bundled runtime secret-store (`cli/lib/runtime/secret-store.js` + launcher wiring) was considered for extraction into its own PR, but per owner decision is dropped together with this PR rather than kept.

Downstream spec (latest-only, zylos-core zero-change): coco-workspace/cws-zylos-runtime#144 (v6). Branch left intact in case any of this is revived later.

